PR #1, opened 2018-12-03
Feature/hana installation

arbulu89:
Updates for the task:
https://trello.com/c/fp52Pvf6/227-create-salt-formula-for-sap-hana-system-replication-and-installation-modules

Add HANA installation features (modules and states)


salt/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
+:author: xarbulu
+:organization: SUSE Linux GmbH
+:contact: xarbulu@suse.com
+
+:since: 2018-11-30
+"""
+
+__version__ = "0.2.0"
```
diegoakechi: Maybe to be coherent with shaptools, we should only increase this version number after the first release...

arbulu89: Agree!

arbulu89:
@diegoakechi Could you accept the PR

salt/modules/hanamod.py
```diff
@@ -98,7 +198,7 @@ def is_running(sid=None, inst=None, password=None):
 
     .. code-block:: bash
 
-        salt '*' hana.is_running prd 00 pass
+        salt '*' hana.is_running prd '"00"' pass
```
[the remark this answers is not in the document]
arbulu89: Yes, this is due yaml parsing, it does strange things. I don't really like even how it works, but by now is better to follow the salt recommendations, and they use this syntax.

PR #3, opened 2019-01-11
Feature/crmsh implementation

arbulu89:
Create salt crmshmod salt module and state code to provide crmsh core functionalities

### DIFF
--- a/salt/states/crmshmod.py
+++ b/salt/states/crmshmod.py
@@ -62,7 +62,7 @@ def cluster_absent(
            'result': False,
            'comment': ''}
 
-    if __salt__['crm.cluster_status']():
+    if __salt__['crm.status']():
         ret['result'] = True
         ret['comment'] = 'Cluster is not running already'
         return ret
@@ -131,7 +131,7 @@ def cluster_initialized(
            'result': False,
            'comment': ''}
 
-    if not __salt__['crm.cluster_status']():
+    if not __salt__['crm.status']():
         ret['result'] = True
         ret['comment'] = 'Cluster is already initialized'
         return ret
@@ -190,7 +190,7 @@ def cluster_joined(
            'result': False,
            'comment': ''}
 
-    if not __salt__['crm.cluster_status']():
+    if not __salt__['crm.status']():
         ret['result'] = True
         ret['comment'] = 'Node is already joined to a cluster'
         return ret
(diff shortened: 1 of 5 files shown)


ayoub-belarbi:
Well done @arbulu89, looks of for me :)

salt/states/crmshmod.py
```diff
@@ -0,0 +1,281 @@
...
+        execute the command in quiet mode (no output)
+    """
+    ret = {'name': name,
+           'changes': {},
+           'result': False,
+           'comment': ''}
+
+    if __salt__['crm.status']():
+        ret['result'] = True
+        ret['comment'] = 'Cluster is not running already'
```
diegoakechi: Maybe I didn't understand the logic here. If the result of `crm status` is true, doesn't it mean that the cluster is still running?
I would also change the message to:

```suggestion
        ret['comment'] = 'Cluster is already not running'
```

salt/states/crmshmod.py
```diff
@@ -0,0 +1,281 @@
...
+        return ret
+
+    if __opts__['test']:
+        ret['result'] = None
+        ret['comment'] = 'Cluster in {} would be removed'.format(name)
```
diegoakechi:
```suggestion
        ret['comment'] = 'Cluster node {} would be removed'.format(name)
```

salt/states/crmshmod.py
```diff
@@ -0,0 +1,281 @@
...
+    try:
+        #  Here starts the actual process
+        result = __salt__['crm.cluster_remove'](
+            host=name,
+            force=True,
+            quiet=quiet)
+
+        if result:
+            ret['changes']['name'] = name
+            ret['comment'] = 'Error removing cluster'
```
diegoakechi:
```suggestion
            ret['comment'] = 'Error removing cluster node'
```

salt/states/crmshmod.py
```diff
@@ -0,0 +1,281 @@
...
+            ret['result'] = False
+            return ret
+
+        ret['changes']['name'] = name
+        ret['comment'] = 'Cluster removed'
```
diegoakechi:
```suggestion
        ret['comment'] = 'Cluster node removed'
```

salt/states/crmshmod.py
```diff
@@ -0,0 +1,281 @@
...
+            the sbd is set as diskless.
+    quiet:
+        execute the command in quiet mode (no output)
+    """
+    ret = {'name': name,
+           'changes': {},
+           'result': False,
+           'comment': ''}
+
+    if not __salt__['crm.status']():
```
diegoakechi: As mentioned above, the logic seems inverted for me for the status call...

salt/states/crmshmod.py
```diff
@@ -0,0 +1,281 @@
...
+        #  Here starts the actual process
+        result = __salt__['crm.cluster_join'](
+            host=name,
+            watchdog=watchdog,
+            interface=interface,
+            quiet=quiet)
+
+        if result:
+            ret['changes']['name'] = name
+            ret['comment'] = 'Error joining to cluster'
```
diegoakechi:
```suggestion
            ret['comment'] = 'Error joining to the cluster'
```

salt/states/crmshmod.py
```diff
@@ -0,0 +1,281 @@
...
+            ret['result'] = False
+            return ret
+
+        ret['changes']['name'] = name
+        ret['comment'] = 'Node joined to cluster'
```
diegoakechi:
```suggestion
        ret['comment'] = 'Node joined to the cluster'
```

salt/states/crmshmod.py
arbulu89: Yeah, maybe it's not totally clear how it works...
Basically, __salt__['crm.status'} returns the returncode of the command crm status. So, if the code is 0, everything went fine, ofherwise some error happened.

We could improve that in future changes anyway

arbulu89:
@diegoakechi I have done the changes. Anyway they are not visible here. As we changed the repo owner, and I did the changes after that in my fork, this PR is not gathering them. What we can do is merge this PR, and after that I could create the new PR from my fork with the changes. What do you think?

diegoakechi:
> @diegoakechi I have done the changes. Anyway they are not visible here. As we changed the repo owner, and I did the changes after that in my fork, this PR is not gathering them. What we can do is merge this PR, and after that I could create the new PR from my fork with the changes. What do you think?

Works for me.

PR #7, opened 2019-02-06
Feature/crmsh implementation

arbulu89:
Update the way crm installation is checked.
Update LICENSE to apache (salt uses apache license)


diegoakechi:
I think the job died during the execution. I have restarted that. Let's see...

PR #12, opened 2019-04-09
Add cost-optimized related functionalities

Simranpal:
1.write wrapper functions for shaptools:
set_ini_parameter
unset_ini_parameter
reduce_memory_resources
reset_memory_parameters

2.define state 'memory_resources_reduced' for reducing HANA memory resources

### DIFF
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -736,10 +736,10 @@ def set_ini_parameter(
     hana_inst = _init(sid, inst, password)
     try:
         hana_inst.set_ini_parameter(
-            ini_parameter_values, database,
-            file_name, layer,
-            layer_name,reconfig,
-            key_name, user_name, user_password)
+            ini_parameter_values=ini_parameter_values,database=database,
+            file_name=file_name, layer=layer,
+            layer_name=layer_name, reconfig=reconfig,
+            key_name=key_name, user_name=user_name, user_password=user_password)
     except hana.HanaError as err:
         raise exceptions.CommandExecutionError(err)
 
@@ -800,9 +800,9 @@ def unset_ini_parameter(
     hana_inst = _init(sid, inst, password)
     try:
         hana_inst.unset_ini_parameter(
-            ini_parameter_names, database,
-            file_name, layer,
-            layer_name, reconfig,
-            key_name, user_name, user_password)
+            ini_parameter_names=ini_parameter_names, database=database,
+            file_name=file_name, layer=layer,
+            layer_name=layer_name, reconfig=reconfig,
+            key_name=key_name, user_name=user_name, user_password=user_password)
     except hana.HanaError as err:
         raise exceptions.CommandExecutionError(err)
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -681,20 +681,27 @@ def sr_cleanup(
 
 
 def set_ini_parameter(
+        ini_parameter_values,
         database,
         file_name,
         layer,
-        parameter_list,
+        layer_name=None,
+        reconfig=False,
+        key_name=None,
+        user_name=None,
+        user_password=None,
         sid=None,
         inst=None,
-        password=None,
-        **kwargs):
+        password=None):
     '''
     Update HANA ini configuration parameter
 
     key_name or user_name/user_password combination,
     one of them must be provided
 
+    ini_parameter_values
+        Dictionary containing HANA parameters & their values
+        where key = ('section_name', 'parameter_name') and value = 'parameter_value'
     database
         Database name
     file_name
@@ -703,8 +710,6 @@ def set_ini_parameter(
         Target layer for the configuration change
     layer_name
         Target either a tenant name or a host name(optional)
-    parameter_list
-        List containing parameter details:section_name, parameter_name, parameter_value
     reconfig
         If apply changes to running HANA instance(optional)
     key_name
@@ -724,41 +729,42 @@ def set_ini_parameter(
 
     .. code-block:: bash
 
-        salt '*' hana.set_ini_parameter key pass SYSTEMDB
-        global.ini HOST node01 row_engine consistency_check_at_startup true
+        salt '*' hana.set_ini_parameter {('memorymanager','global_allocation_limit'): '21000',
+        ('system_replication', 'preload_column_tables'):'False'}
+        SYSTEMDB global.ini HOST node01 key prd '"00"' pass
     '''
-    key_name = kwargs.get('key_name', None)
-    user_name = kwargs.get('user_name', None)
-    user_password = kwargs.get('user_password', None)
-
-    layer_name = kwargs.get('layer_name', None)
-    reconfig = kwargs.get('reconfig', False)
-
     hana_inst = _init(sid, inst, password)
     try:
         hana_inst.set_ini_parameter(
-            database, file_name, layer,
-            parameter_list,
-            layer_name, reconfig,
+            ini_parameter_values, database,
+            file_name, layer,
+            layer_name,reconfig,
             key_name, user_name, user_password)
     except hana.HanaError as err:
         raise exceptions.CommandExecutionError(err)
 
 def unset_ini_parameter(
+        ini_parameter_names,
         database,
         file_name,
         layer,
-        parameter_list,
+        layer_name=None,
+        reconfig=False,
+        key_name=None,
+        user_name=None,
+        user_password=None,
         sid=None,
         inst=None,
-        password=None,
-        **kwargs):
+        password=None):
     '''
     Update HANA ini configuration parameter
 
     key_name or user_name/user_password combination,
     one of them must be provided
 
+    ini_parameter_names: 
+        List of HANA parameter names where each entry looks like
+        (<section_name>,<parameter_name>)
     database
         Database name
     file_name
@@ -767,8 +773,6 @@ def unset_ini_parameter(
         Target layer for the configuration change
     layer_name
         Target either a tenant name or a host name(optional)
-    parameter_list
-        list containing parameter details:section_name, parameter_name
     reconfig
         If apply changes to running HANA instance(optional)
     key_name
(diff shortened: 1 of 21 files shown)


salt/modules/hanamod.py
```diff
@@ -678,3 +678,225 @@ def sr_cleanup(
...
+    '''
+    Update HANA ini configuration parameter
+
+    key_name or user_name/user_password combination,
+    one of them must be provided
+
+    database
+        Database name
+    file_name
+        ini configuration file name
```
arbulu89: Start sentences with upper case

salt/modules/hanamod.py
```diff
@@ -678,3 +678,225 @@ def sr_cleanup(
...
+
+def set_ini_parameter(
+        database,
+        file_name,
+        layer,
+        parameter_list,
+        sid=None,
+        inst=None,
+        password=None,
+        **kwargs):
```
arbulu89: I know that it might sound confusing, but remove the kwargs here and add all the parameters.

In salt code we tend to set all of the parameters to make everything more readable (check other salt modules to get more examples).

salt/modules/hanamod.py
```diff
@@ -678,3 +678,225 @@ def sr_cleanup(
...
+
+def unset_ini_parameter(
+        database,
+        file_name,
+        layer,
+        parameter_list,
+        sid=None,
+        inst=None,
+        password=None,
+        **kwargs):
```
arbulu89: Same than before

salt/modules/hanamod.py
```diff
@@ -678,3 +678,225 @@ def sr_cleanup(
...
+    try:
+        hana_inst.unset_ini_parameter(
+            database, file_name, layer,
+            parameter_list,
+            layer_name,
+            key_name, user_name, user_password)
+    except hana.HanaError as err:
+        raise exceptions.CommandExecutionError(err)
+
+def update_memory_resources(
```
arbulu89: these methods are totally redundant. Do they have any sense? Maybe it should be better to only use set_ini_parameters. In fact, the docstring and the code don't match

salt/modules/hanamod.py
```diff
@@ -678,3 +678,225 @@ def sr_cleanup(
...
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' hana.reset_memory_parameters prd '"00"' pass
+    '''
+    hana_inst = _init(sid, inst, password)
+    try:
+        hana_inst.reset_memory_parameters(
```
arbulu89: This method uses different parameters that the one in shaptools

salt/states/hanamod.py
```diff
@@ -570,3 +570,99 @@ def sr_clean(
...
+        ret['changes']['global_allocation_limit'] = global_allocation_limit
+        return ret
+
+    running = __salt__['hana.is_running'](
+        sid=sid,
+        inst=inst,
+        password=password)
+    #TODO: check existing memory settings
+    
+    parameter_list = {('system_replication', 'preload_column_tables'): preload_column_tables, (\
```
arbulu89: just put ```(``` in the other line

salt/states/hanamod.py
```diff
@@ -570,3 +570,99 @@ def sr_clean(
...
+                parameter_list=parameter_list,
+                key_name=userkey_data.get('key_name'),
+                user_name=userkey_data.get('user_name'),
+                user_password=userkey_data.get('user_password'),
+                sid=sid,
+                inst=inst,
+                password=password)
+            ret['changes']['global_allocation_limit'] = global_allocation_limit
+            #restart HANA for memory changes to take effect
+            if running:
```
arbulu89: We should create a hana.restart method in salt modules code. It would make sense

salt/states/hanamod.py
```diff
@@ -570,3 +570,99 @@ def sr_clean(
...
+            userkey_data = _parse_dict(userkey)
+            #ensure HANA is running for SQL to execute
+            if not running:
+                __salt__['hana.start'](
+                    sid=sid,
+                    inst=inst,
+                    password=password)
+            
+            __salt__['hana.update_memory_resources'](
+                database=userkey_data.get('database', None),
```
arbulu89: Can the database be None? Do we accept that?

salt/states/hanamod.py
```diff
@@ -570,3 +570,99 @@ def sr_clean(
...
+
+    running = __salt__['hana.is_running'](
+        sid=sid,
+        inst=inst,
+        password=password)
+    #TODO: check existing memory settings
+    
+    parameter_list = {('system_replication', 'preload_column_tables'): preload_column_tables, (\
+                       'memorymanager', 'global_allocation_limit') : global_allocation_limit}
+    file_name = 'global.ini'
```
arbulu89: If this variables are constants, we should put them outside this method (at the top of the file) and use upper case to define them

salt/modules/hanamod.py
Simranpal: okay @arbulu89 , i did that.
Also the ini_parameter_values needs to be passed here as dictionary argument

salt/modules/hanamod.py
Simranpal: Yes @arbulu89 , the task can be done via set_ini_parameters, the reason for creation was to use the corresponding fn with hardcoded SQL

salt/states/hanamod.py
Simranpal: yes @arbulu89 , i was thinking same. One question regarding salt actions:
Since __salt__['hana.stop'] module may take some time (depending on HANA shutdown time). and then here I do if not running: __salt__['hana.start'] will this action maybe skipped?
I am thinking because it is one command and not inside some loop. This logic maybe improved, and the logic makes sense to be moved to shaptools.

Simranpal: last doubt/question was cleared, after I test the state in HANA machine. Basically the actions will occur in order according to logic/sequence coded in the state.

diegoakechi:
@Simranpal we have renamed the project. Can you, please, rebase your branch to avoid conflicts?

salt/modules/hanamod.py
Simranpal: redundant fns were removed

salt/states/hanamod.py
Simranpal: I moved the variables to top of file as constants.
These values will not change for the function

salt/states/hanamod.py
Simranpal: I changed the database param to be constant "SYSTEMDB" as it remains same for the related functionality

salt/modules/hanamod.py
```diff
@@ -678,3 +678,130 @@ def sr_cleanup(
...
+        sid=None,
+        inst=None,
+        password=None):
+    '''
+    Update HANA ini configuration parameter
+
+    key_name or user_name/user_password combination,
+    one of them must be provided
+
+    ini_parameter_values
```
arbulu89: The same than in shaptools, I will update the list with the params data by a dictionary

salt/states/hanamod.py
```diff
@@ -67,6 +67,13 @@
 __virtualname__ = 'hana'
 
 TMP_CONFIG_FILE = '/tmp/hana.conf'
+GLOBAL_INI_FILE = 'global.ini'
+HANA_SYSTEMDB = 'SYSTEMDB'
+HANA_SYSTEM_LAYER = 'SYSTEM'
+HANA_LAYER_NAME= None
+HANA_RECONFIG_TRUE= True
+INI_PARAM_PRELOAD_CS = ['system_replication','preload_column_tables']
```
arbulu89: A space missing after the comma

salt/modules/hanamod.py
```diff
@@ -678,3 +678,130 @@ def sr_cleanup(
...
+    hana_inst = _init(sid, inst, password)
+    try:
+        hana_inst.set_ini_parameter(
+            ini_parameter_values, database,
+            file_name, layer,
+            layer_name,reconfig,
+            key_name, user_name, user_password)
+    except hana.HanaError as err:
+        raise exceptions.CommandExecutionError(err)
+
```
arbulu89: 2 new lines in the methods out of the classes

salt/states/hanamod.py
```diff
@@ -562,3 +569,103 @@ def sr_clean(
...
+    user_password
+        Password to connect to sap hana db
+    sid
+        System id of the installed hana platform
+    inst
+        Instance number of the installed hana platform
+    password
+        Password of the installed hana platform user
+    '''
+    INI_PARAM_PRELOAD_CS.append(preload_column_tables)
```
arbulu89: Use dicts as commented before

salt/states/hanamod.py
```diff
@@ -562,3 +569,103 @@ def sr_clean(
...
+        #ensure HANA is running for SQL to execute
+        if not running:
+            __salt__['hana.start'](
+                sid=sid,
+                inst=inst,
+                password=password)
+        
+        __salt__['hana.set_ini_parameter'](
+            ini_parameter_values=ini_parameter_values,
+            database=HANA_SYSTEMDB,
```
arbulu89: Do we really need to use constants this way?
I know that we discussed that in the past, but if those values are going to be used just here, I would put them here, without using global constants.

If the constants are used elsewhere (or the state file must provide them) it is OK

salt/states/hanamod.py
```diff
@@ -562,3 +569,103 @@ def sr_clean(
...
+            reconfig=HANA_RECONFIG_TRUE,
+            user_name=user_name,
+            user_password=user_password,
+            sid=sid,
+            inst=inst,
+            password=password)
+        ret['changes']['global_allocation_limit'] = global_allocation_limit
+        ret['changes']['preload_column_tables'] = preload_column_tables
+        #restart HANA for memory changes to take effect
+        if running:
```
arbulu89: Does this `if` have any sense? You are using the `running` parameter that is already obsolete (you have run `start` before). If you need the actual value run `is_running` again or don't check the `running` variable at all

arbulu89:
Closed as changes are done in other PR

PR #13, opened 2019-04-17
Feature/ha cluster methods

arbulu89:
Depending on the crmsh version we need to use crm or ha-cluster-init/join.

This PR adapts the code to use one or the other depending on the current installed crmsh version.

The rest of the states and formulas are not affected, as the whole management is done in the salt module

### DIFF
--- a/salt-saphana.changes
+++ b/salt-saphana.changes
@@ -2,7 +2,7 @@
 Wed Apr 17 08:28:30 UTC 2019 - Xabier Arbulu Insausti <xarbulu@suse.com>
 
 - Add ha-cluster-init and ha-cluster-join usage when crmsh version
-  is older that 3.0.0 
+  is older than 3.0.0 
 
 -------------------------------------------------------------------
 Mon Mar  4 15:17:24 UTC 2019 - xarbulu@suse.com
@@ -11,7 +11,7 @@ Mon Mar  4 15:17:24 UTC 2019 - xarbulu@suse.com
   the user_name/user_password is not needed.
 - Removed salt warnings because of non usage of 'name' parameter as
   the first parameter of the state.
-     
+
 -------------------------------------------------------------------
 Mon Feb  25 10:22:24 UTC 2019 - dakechi@suse.com
 
@@ -21,7 +21,7 @@ Mon Feb  25 10:22:24 UTC 2019 - dakechi@suse.com
 -------------------------------------------------------------------
 Wed Feb  6 14:09:24 UTC 2019 - xarbulu@suse.com
 
-- Update the way crm installation is checked 
+- Update the way crm installation is checked
 
 -------------------------------------------------------------------
 Thu Jan 17 13:24:33 UTC 2019 - xarbulu@suse.com
@@ -31,9 +31,9 @@ Thu Jan 17 13:24:33 UTC 2019 - xarbulu@suse.com
 -------------------------------------------------------------------
 Fri Jan 11 08:22:59 UTC 2019 - xarbulu@suse.com
 
-- Create crmsh core functionalities salt module and states code 
+- Create crmsh core functionalities salt module and states code
 
 -------------------------------------------------------------------
 Thu Dec 20 08:33:10 UTC 2018 - xarbulu@suse.com
 
-- First package version 
+- First package version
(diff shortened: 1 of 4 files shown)


salt-saphana.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Apr 17 08:28:30 UTC 2019 - Xabier Arbulu Insausti <xarbulu@suse.com>
+
+- Add ha-cluster-init and ha-cluster-join usage when crmsh version
+  is older that 3.0.0
```
diegoakechi:
``` Suggestion
  is older than 3.0.0 
```

salt/modules/crmshmod.py
```diff
@@ -192,20 +281,47 @@ def cluster_init(
...
     '''
-    cmd = '{crm_command} cluster init -y -n {name}'.format(
-        crm_command=CRM_COMMAND, name=name)
+    # INFO: 2 different methods are created to make easy to read/understand
+    # and create the corresponing UT
+    if __crmnewversion__:
+        return _crm_init(
+            name, watchdog, interface, unicast, admin_ip, sbd, sbd_dev, quiet)
+
+    LOGGER.warn('The parameters name and watchdog are not considered!')
```
diegoakechi: @arbulu89 I think that here, for the older versions, we will need to load the watchdog module manually, to abstract this step to the end user. Even that the ha-cluster scripts didn't cover that at this time, the watchdog configuration was needed. What do you think @krig?

arbulu89: @diegoakechi Yes, actually we do this in habootstrap-formula.

I didn't want to add this extra logic here

arbulu89: @diegoakechi What do we have to do for that?
The watchdog module is loaded at least in habootstrap-formula, I don't know if we have to say to the cluster which of the watchdogs to use

diegoakechi: You need to point SBD what watchdog to use. On the `/etc/sysconfig/sbd` this parameter does that:
`SBD_WATCHDOG_DEV=/dev/watchdog`

If the formula is doing both, that is fine. If not, we need to do it. I think there is a SBD option for that.

arbulu89: @diegoakechi I will have a look! Ther formula is not doing that, only loading the watchdog. I will need to add this extra step here.

diegoakechi: You need to configure this before creating the SBD device. So, it should be configured before the sbd create command. @gao-yan Can you confirm that?

krig: I am not sure `SBD_WATCHDOG_DEV` existed in earlier versions of sbd, might need to look into that.

salt/modules/crmshmod.py
```diff
@@ -154,6 +182,66 @@ def wait_for_startup(
...
+    '''
+    ha-cluster-init command execution
+    '''
+    cmd = '{ha_init_command} -y'.format(
+        ha_init_command=HA_INIT_COMMAND)
+    if interface:
+        cmd = '{cmd} -i {interface}'.format(cmd=cmd, interface=interface)
+    if unicast:
+        cmd = '{cmd} -u'.format(cmd=cmd)
+    if admin_ip:
```
krig: Older versions of `ha-cluster-init` may lack some of these parameters.. In particular, I think `-S` didn't exist before the move to `crm cluster init`.. but `-u` and `-A` might also be newer.

arbulu89: I have tested them and in sle12sp1 and sle12sp2 they work, at least `ha-cluster-init` doesn't complain about bad options. But yes, this doesn't mean that they do what they are supposed to do

diegoakechi: Well, some tests will be needed to check if the options do what they should do... =)

arbulu89:
@diegoakechi @krig I have fount that I need to create more code to manipulate corosync configuration. Old ha-cluster-init package is not setting unicast, so we have to do it manually. I will add some code to do that, test and we will review later again.

diegoakechi:
@arbulu89 please, rebase to avoid conflicts with the changes we did for renaming the project.

diegoakechi:
LGTM.

PR #15, opened 2019-04-23
Remove python shaptools dependency from spec file

arbulu89:
As python shaptools is the module used in the minions
(or local if executed as minion) it shouldn't be
a hard requirement

### DIFF
--- a/salt-saphana.changes
+++ b/salt-saphana.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Tue Apr 23 14:40:53 UTC 2019 - Xabier Arbulu Insausti <xarbulu@suse.com>
+
+- Remove python shaptools dependency from spec file 
+
+-------------------------------------------------------------------
 Mon Mar  4 15:17:24 UTC 2019 - xarbulu@suse.com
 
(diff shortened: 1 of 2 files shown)


diegoakechi:
@arbulu89 in that case, I think we should change it to `Suggests` as python-shaptools is still needed, and then, if the user wants, he can install the suggested package.

arbulu89:
> Suggested a change.

 In my head now we shouldn't include anything about shaptools there hehe. That's why: when we install salt-saphana we are like installing salt (actually remember that this code should/will go in salt upstream). That means that python shaptools must only be installed in the minions and not neccessarily in the master. So when we install salt-saphana we shouldn't install shaptools, this should be done by the user and might be done using salt for example. That's what our formulas are doing.
Most of the salt modules depends on other python packages, but they are not required during salt installation, it wouldn't make sense. I see our example as the same
Actually, salt would raise an internal error saying `crmshmod not loaded`. This happens for example if you try to use the dockermod but you don't have the docker-py module installed

arbulu89:
Changes applied in:
https://github.com/SUSE/salt-saphana/pull/18

PR #18, opened 2019-04-25
Rename the RPM package from salt-saphana to salt-shaptools

diegoakechi:
This package is intended to host SAP Applications and SLE-HA salt modules and states, so a more meaningful name is needed.


diegoakechi:
> There are some references in README.md to salt-saphana, we should change them too (mainly the badges).
The Code Climate includes a Hash in the link, so I will do it after renaming the repo and reconfiguring the integrations.

> 
> Besides that, we should maybe set the version back to 0.1.0 (i don't know why i increased that...)
Done.

> Finally, in order to make the CI/CD, we need to create the projects in the OBS repositories (empty packages).

This was already done.

> Finally, even though we have other PR for that, I would take advantage of removing the python-shaptools depedency from the spec file
Done.
> (we can remove the `%define fname hana` line too, it's not used)

Done.

PR #24, opened 2019-05-08
Functionality to reduce HANA memory resources

Simranpal:
I created a new PR with fewer commit history and all the relevant code changes.
This includes:
1) wrapper module functions for shaptools:
-set_ini_parameter
-unset_ini_parameter

2) salt state 'memory_resources_reduced' for reducing HANA memory resources

### DIFF
--- a/salt/states/hanamod.py
+++ b/salt/states/hanamod.py
@@ -67,6 +67,8 @@ from salt.ext import six
 __virtualname__ = 'hana'
 
 TMP_CONFIG_FILE = '/tmp/hana.conf'
+INI_PARAM_PRELOAD_CS = {'section_name':'system_replication', 'parameter_name':'preload_column_tables'}
+INI_PARAM_GAL = {'section_name':'memorymanager', 'parameter_name':'global_allocation_limit'}
 
 
 def __virtual__():  # pragma: no cover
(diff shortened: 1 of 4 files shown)


Simranpal:
Hi @arbulu89 , I have added the UT for corresponding functions with 100% coverage now

salt/states/hanamod.py
```diff
@@ -67,6 +67,8 @@
 __virtualname__ = 'hana'
 
 TMP_CONFIG_FILE = '/tmp/hana.conf'
+INI_PARAM_PRELOAD_CS = {'section_name':'system_replication', 'parameter_name':'preload_column_tables'}
```
arbulu89: Are these lines shorter than 100 characters?
Please check that?
I guess you are using a linter in your code editor, but just in case

salt/states/hanamod.py
```diff
@@ -562,3 +564,103 @@ def sr_clean(
...
+    user_password
+        Password to connect to sap hana db
+    sid
+        System id of the installed hana platform
+    inst
+        Instance number of the installed hana platform
+    password
+        Password of the installed hana platform user
+    '''
+    INI_PARAM_PRELOAD_CS['parameter_value'] = preload_column_tables
```
arbulu89: To update constants on the run is quite a bad practice.
But I guess we can leave this way by now

PR #31, opened 2019-08-02
Add initial skeleton and code for NW salt code

arbulu89:
Firs implementation of the Netweaver salt module.

The scope of this PR is to have the minimum modules and methods to install the ASCS SAP instance

### DIFF
--- a/salt/modules/netweavermod.py
+++ b/salt/modules/netweavermod.py
@@ -104,6 +104,40 @@ def is_installed(
     return netweaver_inst.is_installed(sap_instance)
 
 
+def attach_virtual_host(
+        virtual_host,
+        virtual_host_interface='eth0'):
+    '''
+    Attach virtual host ip address to network interface
+
+    virtual_host
+        Virtual host name
+    virtual_host_interface:
+        Network interface to attach the virtual host ip address
+
+    Returns:
+        str: Attached ip address
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' netweaver.attach_virtual_host my_host eth1
+    '''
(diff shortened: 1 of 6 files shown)


salt/modules/netweavermod.py
```diff
@@ -0,0 +1,140 @@
...
+    '''
+    Check if SAP Netweaver platform is installed
+
+    sid
+        Netweaver system id (PRD for example)
+    inst
+        Netweaver instance number (00 for example)
+    password
+        Netweaver instance password
+    sap_intance
```
diegoakechi:
```suggestion
    sap_instance
```

salt/states/netweavermod.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _get_sap_instance_type(product_id):
+    '''
+    Get SAP instance type from the product id string
+    '''
+    return product_id.split(':')[0].split('-')[-1].lower()
+
+
+def installed(
```
diegoakechi: @arbulu89 Wouldn't you change this to `present`?

arbulu89: I would like... But i don't want to have `present` here and `installed` in HANA.
I think we will need to do the change in all of them together.

Anyway, what do you prefer @diegoakechi @Simranpal ? `installed/uninstalled` vs `present/absent`.
The 2nd follows more the salt style in my opinion

diegoakechi: I prefer to follow the salt convention, but I am ok to do it in a special PR changing all at the same time.

Simranpal: I think installed/uninstalled is consistent with HANA

PR #33, opened 2019-09-05
Feature/netweaver ers

arbulu89:
Netweaver ERS installation changes. Mainly based in:
https://github.com/SUSE/shaptools/pull/42


salt/states/netweavermod.py
```diff
@@ -69,7 +69,10 @@ def installed(
         config_file,
         virtual_host,
         virtual_host_interface,
-        product_id):
+        product_id,
+        ascs_password=None,
+        timeout=0,
+        interval=5):
     """
     Install SAP Netweaver if the instance is not installed yet.
```
Simranpal: Here and also in other docstrings we may write 'SAP Netweaver instance' instead of "SAP Netweaver"
or specify the type of instance in certain cases

arbulu89:
@Simranpal, please review the latest changes

PR #34, opened 2019-09-18
Feature/db pas instances

arbulu89:
Add new salt modules and states to manage DB, PAS and AAS Netweaver installations.

To use them, we need to check if a Hana database is available (`wait_for_connection`).
The db installation uses this way to check if the Netweaver DB import is done or not

### DIFF
--- a/salt/modules/netweavermod.py
+++ b/salt/modules/netweavermod.py
@@ -89,7 +89,7 @@ def is_installed(
         Netweaver instance password
     sap_instance
         Check for specific SAP instances. Available options: ascs, ers, pas, aas.
-        If None it will checked if any instance is installed
+        If None it will check if any instance is installed
 
     Returns:
         bool: True if installed, False otherwise
@@ -132,7 +132,7 @@ def is_db_installed(
     '''
     if 'hana.wait_for_connection' not in __salt__:
         raise exceptions.CommandExecutionError(
-            'hana.wait_for_connection not available. hanamod must be installed')
+            'hana.wait_for_connection not available. hanamod must be installed and loaded')
 
     try:
         __salt__['hana.wait_for_connection'](
@@ -206,6 +206,8 @@ def install(
         Root user password
     cwd
         New value for SAPINST_CWD parameter
+        CAUTION: All of the files stored in this path will be removed except the
+        start_dir.cd. This folder only will contain temporary files about the installation
 
     CLI Example:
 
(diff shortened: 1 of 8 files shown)


salt/modules/netweavermod.py
```diff
@@ -88,8 +88,8 @@ def is_installed(
     password
         Netweaver instance password
     sap_instance
-        Check for specific SAP instances. Available options: ascs, ers. If None it will checked
-        if any instance is installed
+        Check for specific SAP instances. Available options: ascs, ers, pas, aas.
+        If None it will checked if any instance is installed
```
diegoakechi:
```suggestion
        If None it will check if any instance is installed
```

salt/modules/netweavermod.py
```diff
@@ -216,6 +265,38 @@ def install_ers(
...
+def setup_cwd(
+        software_path,
+        cwd='/tmp/unattended',
+        additional_dvds=None):
+    '''
+    Setup folder to run the sapinst tool in other directory (modified SAPINST_CWD)
+
+    software_path
+        Path where SAP Netweaver software is downloaded
+    cwd
```
diegoakechi: @arbulu89 I think same warning about the deletion commented on `shaptools` should be included here...

salt/modules/netweavermod.py
```diff
@@ -104,6 +104,49 @@ def is_installed(
...
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' netweaver.is_db_installed 192.168.10.15 30013 SAPABAP1 password
+    '''
+    if 'hana.wait_for_connection' not in __salt__:
+        raise exceptions.CommandExecutionError(
+            'hana.wait_for_connection not available. hanamod must be installed')
```
MalloZup: nitpick
I think this is half-away correct.
https://docs.saltstack.com/en/latest/ref/modules/#cross-calling-execution-modules

We should perhaps error-out with `module X not loaded in minion. Hanamod must be installed or loaded.`

The if condition is just checking that the module is loaded on minion which isn't same as installed

salt/states/netweavermod.py
```diff
@@ -136,12 +143,25 @@ def installed(
             virtual_host=virtual_host,
             virtual_host_interface=virtual_host_interface)
 
+        if cwd:
+            cwd = __salt__['netweaver.setup_cwd'](
```
MalloZup: are we sure that we don't need to check error here? with a try catch?

salt/modules/netweavermod.py
MalloZup: as motivation is that we might encounter situation where the module was not loaded on minion for X reason

tests/unit/states/test_hanamod.py
```diff
@@ -31,6 +31,70 @@ class HanamodTestCase(TestCase, LoaderModuleMockMixin):
     def setup_loader_modules(self):
         return {hanamod: {'__opts__': {'test': False}}}
 
+    # 'available' function tests
+
```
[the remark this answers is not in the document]
arbulu89: Yes, just to follow the same style than others

PR #35, opened 2019-09-26
Feature/ha netweaver

arbulu89:
New salt methods and states used in the last PR of `sapnewbootstrap-formula`:
https://github.com/SUSE/sapnwbootstrap-formula/pull/11

As this last PR, the methods are implemented to enable HA in ASCS and ERS

### DIFF
--- a/salt/states/netweavermod.py
+++ b/salt/states/netweavermod.py
@@ -304,3 +304,147 @@ def db_installed(
     except exceptions.CommandExecutionError as err:
         ret['comment'] = six.text_type(err)
         return ret
+
+
+def check_instance_present(
+            name,
+            dispstatus=None,
+            virtual_host=None,
+            sid=None,
+            inst=None,
+            password=None):
+        '''
+        Check if a SAP Netweaver instance is installed among all the nodes
+
+        name:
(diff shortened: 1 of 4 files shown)


salt/states/netweavermod.py
```diff
@@ -304,3 +304,147 @@ def db_installed(
...
+        ret['comment'] = 'invalid sap_instance. Only \'ascs\' and \'ers\' are valid options'
+        return ret
+
+    if not __salt__['cmd.retcode']('cat {} | grep \'.*{}.*\''.format(
+            sapservice_file, 'ASCS' if sap_instance == 'ers' else 'ERS'), python_shell=True):
+        ret['result'] = True
+        ret['comment'] = 'sapservices file is already updated for instance {}'.format(sap_instance)
+        return ret
+
+    if __opts__['test']:
```
MalloZup: `nit` I see lot of if/else/if. Perhaps we should add a comment to explain the different behaviour they are adressing on the highlevel manner

arbulu89: All of the `if/else` are quite easy to follow as all of them are eventually used to decide which comment is going to be set for the return

PR #37, opened 2019-10-28
Rename drbd.py to drbdmod.py to unified style.

nick-wang:
Change the drbd file name to drbdmod


nick-wang:
@arbulu89 I don't have write access to merge, so need your help on this. Thanks for the review.

Issue #39, opened 2019-11-19
Error parsing drbd status


arbulu89:
This error might be raised during `drbd status` command:
```
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec): [INFO    ] Executing command 'drbdadm status sapdata' in directory '/root'
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec): [ERROR   ] An exception occurred in this state: Traceback (most recent call last):
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):   File "/usr/lib/python3.6/site-packages/salt/state.py", line 1939, in call
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):     ret = self.states[cdata['full']](*cdata['args'], **cdata['kwargs'])
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):   File "/usr/lib/python3.6/site-packages/salt/loader.py", line 1982, in wrapper
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):     return f(*args, **kwargs)
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):   File "/var/cache/salt/minion/extmods/states/drbdmod.py", line 450, in wait_for_successful_synced
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):     **kwargs)
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):   File "/var/cache/salt/minion/extmods/modules/drbdmod.py", line 639, in check_sync_status
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):     if _is_local_all_uptodated(name) and _is_peers_uptodated(
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):   File "/var/cache/salt/minion/extmods/modules/drbdmod.py", line 210, in _is_local_all_uptodated
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):     res = status(name)
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):   File "/var/cache/salt/minion/extmods/modules/drbdmod.py", line 372, in status
module.drbd_node.null_resource.drbd_node_provisioner[0] (remote-exec):     __context__['drbd.statusret'].append(__context__['drbd.resource'])
[...]

PR #40, opened 2019-11-21
Fix Issue39 DRBD error parsing status

nick-wang:
Issue: https://github.com/SUSE/salt-shaptools/issues/39
    
The `drbdadm status` output may to new line with small resolution.
Especeially when syncing congested.
Ignore the indent 6 and 8 in this case.
    
Eventually will replace the whole method with `json` output.

### DIFF
--- a/salt/modules/drbdmod.py
+++ b/salt/modules/drbdmod.py
@@ -30,6 +30,8 @@ LOGGER = logging.getLogger(__name__)
 __virtualname__ = 'drbd'
 
 DRBD_COMMAND = 'drbdadm'
+ERR_STR = 'UNKNOWN'
+DUMMY_STR = 'IGNORED'
 
 
 def __virtual__():  # pragma: no cover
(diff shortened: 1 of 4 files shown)


salt/modules/drbdmod.py
```diff
@@ -365,8 +372,13 @@ def status(name='all'):
         LOGGER.info('No status due to %s (%s).', result['stderr'], result['retcode'])
         return None
 
-    for line in result['stdout'].splitlines():
-        _line_parser(line)
+    try:
+        for line in result['stdout'].splitlines():
+            _line_parser(line)
+    except CommandExecutionError as err:
```
arbulu89: Do you think that we need to re raise the error here?
I mean, it catches the error and raises the same one with a small change in the text.
I think we could spare this `try/catch` here

salt/modules/drbdmod.py
```diff
@@ -98,6 +100,9 @@ def _analyse_status_type(line):
         if x in line:
             return ret[x]
 
+    # Doesn't find expected KEY in support indent
+    return 'UNKNOWN'
```
arbulu89: I would put this `UNKNOWN` in a constant, but, it's OK this way too.

salt/modules/drbdmod.py
nick-wang: It is because only the `line` that parser doesn't know can be shown in the _line_parser(). It will be easier to debug with the whole output `result['stdout']`. Result is like
```
Error running 'drbd.status': UNKNOWN status output format found. Additional info follows:

sapdata role:Primary
  disk:UpToDate
  libvirt-sles15-false-po-drbd02 role:Secondary congested:yes ap-in-flight:0
      rs-in-flight:8192
    replication:SyncSource peer-disk:Inconsistent done:21.22

Unknown line:
      rs-in-flight:8192
```

arbulu89: Ok, it makes sense

nick-wang:
@arbulu89 I fixed the constant, though it may be replaced soon.
With the hot fix, the original error caused by things like `rs-in-flight:8192` will be ignored.

But the more i fix it, the more i missing the `json` way. Will change it soon!

PR #41, opened 2019-11-28
Change status to json

nick-wang:


nick-wang:
This pull request is PASS the testing of formulas against SLE15SP1. The implementation in unit test is working in progress.

salt/modules/drbdmod.py
```diff
@@ -32,6 +32,7 @@
 DRBD_COMMAND = 'drbdadm'
 ERR_STR = 'UNKNOWN'
 DUMMY_STR = 'IGNORED'
+WITH_JSON = True
```
arbulu89: I don't really understand this. What do you use this constant for? It's always `True`, isn't it?

salt/modules/drbdmod.py
```diff
@@ -216,13 +217,21 @@ def _is_local_all_uptodated(name):
     Check whether all local volumes are UpToDate.
     '''
 
-    res = status(name)
+    if is_support_status_json():
+        volkey = 'devices'
```
arbulu89: I would create a dictionary with these constants with two entries instead of having independent variables: `json_format` and `text_format`.

Maybe the status would be the unique exception.

And depending on the usage type, use one entry data or another. It looks neater.
What do you think?

salt/modules/drbdmod.py
```diff
@@ -655,3 +669,22 @@ def check_sync_status(name, peernode='all'):
         return True
 
     return False
+
+
+def is_support_status_json():
```
arbulu89: The method name looks a bit strange, right?
I would suggest `is_json_format_available`.
But anyway, up to you @nick-wang hehe

salt/modules/drbdmod.py
arbulu89: Now I see that you use it for the Unit Tests. A bit strange.
After seeing that, I think this shouldn't be a constant, It should be a variable, and it should be retrieved depending on the drbd version (json format available or not)

salt/modules/drbdmod.py
```diff
@@ -655,3 +669,22 @@ def check_sync_status(name, peernode='all'):
...
+    '''
+    Use drbdsetup status --json instead of drbdadm for status.
+    At least in drbd9 and drbd-utils >= 8.9.8
+    '''
+
+    if not WITH_JSON:
+        return False
+
+    cmd = 'drbdsetup status --help 2>/dev/null'
```
arbulu89: It wouldn't be better to check the drbd version instead of parsing the `help` output. Just wondering...

salt/states/drbdmod.py
```diff
@@ -288,7 +315,8 @@ def promoted(name, force=False):
     # Check resource is running
     res = _get_res_status(name)
     if res:
-        if res['local role'] == 'Primary':
+        if (__salt__['drbd.is_support_status_json'] and res['role'] == 'Primary') or \
```
arbulu89: Instead of calling this method everytime, could be just store the value of `drbd.is_support_status_json` at the beginning? Otherwise we will be calling all the time this command without really needing

arbulu89: In fact you are duplicating the call. We could just call it at least just once before the if

salt/modules/drbdmod.py
arbulu89: Something like:
```
OUTPUT_OPTIONS = {
"json":{
"volkey":"devices",
"statekey":"disk-state",
"res":setup_status
},
"text":{
"volkey":"local volumes",
"statekey":"disk",
"res":status
}
}
```

salt/modules/drbdmod.py
```diff
@@ -236,19 +245,29 @@ def _is_peers_uptodated(name, peernode='all'):
 
         If peernode is not match, will return None, same as False.
     '''
-    ret = None
+    if is_support_status_json():
+        conkey = 'connections'
```
arbulu89: The same than before

salt/modules/drbdmod.py
```diff
@@ -655,3 +669,22 @@ def check_sync_status(name, peernode='all'):
...
     return False
+
+
+def is_support_status_json():
+    '''
+    Use drbdsetup status --json instead of drbdadm for status.
+    At least in drbd9 and drbd-utils >= 8.9.8
+    '''
+
+    if not WITH_JSON:
```
arbulu89: Is this for the Unit tests? We could do it better I think, just patching the output of this command

salt/modules/drbdmod.py
nick-wang: Actually i add this before creating unit test, which just utilize this.(Or i can mask the command.)

Why i leave this is just in case `drbdsetup json` format changed in the future version and effect by a version upgrade, since normally admin only check the output for `drbdadm` and leave the `json` to script. It is likely i ignore the change and release a new update which break the installation. 
So i would leave a switch as the benefit of implementing two method (`drbdsetup json` and `drbdadm`). Is that make sense?

salt/modules/drbdmod.py
nick-wang: Like above comment, originally is not for unit test.

salt/modules/drbdmod.py
nick-wang: Fixed in the new commit.
Defined OUTPUT_OPTIONS in the middle since need to define after function.

salt/modules/drbdmod.py
nick-wang: Changed in the new commit.

salt/modules/drbdmod.py
nick-wang: Changed in the new commit.

salt/states/drbdmod.py
nick-wang: Fixed in the new commit.

salt/modules/drbdmod.py
nick-wang: Fixed in the new commit.

nick-wang:
@arbulu89 Fixed all except the `WITH_JSON` flag in the new commit.

I don't intent to use it in unit test only, also want a insurance(since already implemented) for a quick convert in case `json` format changed in future without noticed.

salt/modules/drbdmod.py
```diff
@@ -215,14 +216,19 @@ def _is_local_all_uptodated(name):
     '''
     Check whether all local volumes are UpToDate.
     '''
+    if is_json_format_available():
+        format = 'json'
```
arbulu89: Carefull @nick-wang , `format` is a `built-in` method, and here you are replacing it. Choose any other name XD

salt/modules/drbdmod.py
```diff
@@ -236,19 +242,23 @@ def _is_peers_uptodated(name, peernode='all'):
 
         If peernode is not match, will return None, same as False.
     '''
-    ret = None
+    if is_json_format_available():
+        format = 'json'
```
arbulu89: The same than previous comment

salt/modules/drbdmod.py
```diff
@@ -632,6 +649,29 @@ def setup_status(name='all'):
     return ret
 
 
+# Define OUTPUT_OPTIONS after setup_status() and status() defined
+OUTPUT_OPTIONS = {
```
arbulu89: I would put this constant in the top of the file. It's easier to find it. But put it in the place you most like

salt/states/drbdmod.py
```diff
@@ -285,10 +312,13 @@ def promoted(name, force=False):
...
         return ret
 
+    json_format = __salt__['drbd.is_json_format_available']
+
     # Check resource is running
     res = _get_res_status(name)
     if res:
-        if res['local role'] == 'Primary':
+        if (json_format and res['role'] == 'Primary') or \
+            (not json_format and res['local role'] == 'Primary'):
```
arbulu89: Is the indentation of this line correct? It looks like pep8 would complain XD

salt/states/drbdmod.py
```diff
@@ -345,10 +375,13 @@ def demoted(name):
...
         return ret
 
+    json_format = __salt__['drbd.is_json_format_available']
+
     # Check resource is running
     res = _get_res_status(name)
     if res:
-        if res['local role'] == 'Secondary':
+        if (json_format and res['role'] == 'Secondary') or \
+            (not json_format and res['local role'] == 'Secondary'):
```
arbulu89: Indentation... The same as before

salt/states/drbdmod.py
```diff
@@ -285,10 +312,13 @@ def promoted(name, force=False):
         ret['comment'] = 'Resource {} not defined in your config.'.format(name)
         return ret
 
+    json_format = __salt__['drbd.is_json_format_available']
+
     # Check resource is running
     res = _get_res_status(name)
     if res:
-        if res['local role'] == 'Primary':
+        if (json_format and res['role'] == 'Primary') or \
```
arbulu89: I think you could follow the same implementation using `OUTPUT_OPTIONS` here. It would make total sense, and the `if` statement would be much simpler

salt/modules/drbdmod.py
```diff
@@ -655,3 +695,31 @@ def check_sync_status(name, peernode='all'):
...
+
+    # Output of `drbdadm -V`:
+    # DRBDADM_BUILDTAG=GIT-hash:\ xxx\ reproducible\ build\,\ 2019-09-20\ 12:00:00
+    # DRBDADM_API_VERSION=2
+    # DRBD_KERNEL_VERSION_CODE=0x090010
+    # DRBD_KERNEL_VERSION=9.0.16
+    # DRBDADM_VERSION_CODE=0x090600
+    # DRBDADM_VERSION=9.6.0
+
+    cmd = 'drbdadm -V 2>/dev/null'
```
arbulu89: I don't think you need to reinvent the wheel to check the package version. Salt provides good tools for that.
Here the `crmshmod` example:
https://github.com/SUSE/salt-shaptools/blob/master/salt/modules/crmshmod.py#L49

salt/modules/drbdmod.py
arbulu89: Ok, I personally don't like it but if you think it's needed we can keep it

salt/states/drbdmod.py
nick-wang: Fixed in the new commit.

salt/states/drbdmod.py
nick-wang: Fixed in the new commit.

salt/states/drbdmod.py
nick-wang: Fixed in the new commit.

salt/modules/drbdmod.py
nick-wang: Fixed in the new commit

salt/modules/drbdmod.py
nick-wang: Discussed offline. NameError raised when put this on the top, before status() and setup_status() defined.

salt/modules/drbdmod.py
nick-wang: Fixed in the new commit with salt built in version_cmp

salt/modules/drbdmod.py
nick-wang: `WITH_JSON` is not using for result of version compare plus the switch.

nick-wang:
> Hi @nick-wang ,
> Now the code looks great! Ready to merge from my side (run some tests with terraform just in case).
@arbulu89 I did some test with DRBD formula, not with terraform yet.I can do it today, but I think it should be ok since it is all hide in formula. But i will leave the merge until we both tested.

> PD: I appreciate you added all of this asserts for the logging in `crmshmod`, but I think I will remove this output eventually, as it's printed all the time, and is too verbose.
Sorry, i didn't get this. I didn't add new LOGGER, instead i just remove it in `__virtual__` function related to `crmsh` to `crm` you mentioned yesterday.

PR #42, opened 2019-12-11
Feature/no overwrite sshkeys

arbulu89:
Add `--no-overwrite-sshkey` option.
~We cannot merge this PR until the feature in crmsh is backported from SLE12SP3 and beyond~

~This feature is available in all version except SLE12SP3. So, it's kind of secure to merge.
It won't work in SP3, but well, it's just a side effect.~

Ok, editing again. I was wrong, this codebase is not still in the SLE channels, so we need to wait for that before merging this PR


salt/modules/crmshmod.py
```diff
@@ -388,9 +395,10 @@ def cluster_init(
     # and create the corresponing UT
     if __salt__['crm.version']:
         return _crm_init(
-            name, watchdog, interface, unicast, admin_ip, sbd, sbd_dev, quiet)
+            name, watchdog, interface, unicast, admin_ip, sbd, sbd_dev, no_overwrite_sshkey, quiet)
 
     LOGGER.warn('The parameter name is not considered!')
+    LOGGER.warn('--no_overwrite_sshkey option not available')
```
nick-wang: IIRC, `ha-bootstrap-init/join` merged into `crm cluster init/join` since SLE12SP3, so i guess the codes for `_ha_cluster_init` related will eventually dropped.

arbulu89:
> Maybe better to compare `crmsh` version to see whether `no_overwrite_sshkey` is supported or not.
> 
> But considering `crmsh` didn't bump up version with the fix and versioning is slightly mess between SLE12~SLE15. I think it is fine for the current change.
> 
> Good and important change!

Yes, the difference is a commit hash, and I don't want to start checking this versions hehe
So far, the best is just to wait crmsh is backported correctly to avoid more workarounds...

PR #43, opened 2019-12-11
Add the option to use multiple disk with sbd

arbulu89:
New information!
Multiple sbd disks usage is not supported yet by `crmsh`. In order to make the formula work, I have created a workaround while we do the changes in crmsh.

This is more or less:
- If 1 disk is used it works like before
- If more than 1 disk is used:
   - sbd is executed first `sbd -d disk1 -d disk2 create`
   - `crm cluster init sbd -d {disk}' is executed first to initialize sbd configuration files, csync, etc
   - `/etc/sysconfig/sbd` is updated to add all the disks
   - crmsh is executed without sbd options as this is already on place

This is the state after running the formula
```
   CGroup: /system.slice/sbd.service
           ├─8533 sbd: inquisitor
           ├─8551 sbd: watcher: /dev/sda - slot: 0 - uuid: 66c01dba-aa00-46c5-83ff-a44c31065438
           ├─8552 sbd: watcher: /dev/sdb - slot: 0 - uuid: 96806c1a-7ef6-44dc-b8d0-a401323713e0
           ├─8553 sbd: watcher: /dev/sdc - slot: 0 - uuid: 7e88f747-7a85-4c06-8202-ed6a34d5b13c
           ├─8554 sbd: watcher: Pacemaker
           └─8555 sbd: watcher: Cluster
```


arbulu89:
@diegoakechi @Simranpal I added a workaround while crmsh doesn't support multiple disks. It works fine

PR #45, opened 2019-12-17
Add saptune functionality for solution apply

MalloZup:
# Things to do:

- [x] Version check of saptune


salt/states/saptune.py
```diff
@@ -0,0 +1,85 @@
...
+    if __opts__['test']:
+        ret['result'] = None
+        ret['comment'] = 'Saptune {} solution would be removed'.format(name)
+        ret['changes']['name'] = name
+        return ret
+
+    try:
+        #  Here starts the actual process
+        result = __salt__['saptune.revert_solution'](
+            solution_name=name,
```
nick-wang: Looks like it is incomplete.

salt/states/saptune.py
```diff
@@ -0,0 +1,85 @@
...
+    Revert a saptune solution
+    """
+    ret = {'name': name,
+           'changes': {},
+           'result': False,
+           'comment': ''}
+
+    if __salt__['saptune.revert_solution']():
+        ret['result'] = True
+        ret['comment'] = 'Saptune solution is already reverted'
```
nick-wang: Maybe need to find another way to check the solution is not applied. Calling the salt module will actually do the revert `'{0} solution revert {1}'.format(SAPTUNE_BIN, solution_name)`

MalloZup: thx @angelabriel Doing that approach right now

MalloZup:
@arbulu89 from you pov a part of the unit-tests for the states, is something else I'm missing here? tia

salt/modules/saptune.py
```diff
@@ -0,0 +1,60 @@
...
+    Tune system for all notes applicable to your SAP solution:
+
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' saptune.apply_solution solution-name
+    '''
+    cmd = '{0} solution apply {1}'.format(SAPTUNE_BIN, solution_name)
```
arbulu89: I would use some more meaningful names rather than `{0}`.

salt/states/saptune.py
```diff
@@ -0,0 +1,80 @@
...
+
+.. versionadded:: pending
+
+:maintainer:    Dario Maiocchi <dmaiocchi@suse.com>
+:maturity:      alpha
+:depends:       saptune
+:platform:      all
+
+:configuration: This module requires the saptune python module and uses the
+    following defaults which may be overridden in the minion configuration:
```
arbulu89: I would just put here the first sentence `This module requires the saptune module`,
and remove the `code-block` entry. This is used is the state has some pre configuration option using the minion configuration file, but this is not the case

salt/states/saptune.py
```diff
@@ -0,0 +1,80 @@
...
+
+.. code-block:: yaml
+
+:usage:
+
+.. code-block:: yaml
+  solution_is_removed:
```
arbulu89: Maybe `solution_is_applied`

salt/states/saptune.py
```diff
@@ -0,0 +1,80 @@
...
+def solution_applied(name):
+    """
+    Apply a saptune solution
+    """
+    ret = {'name': name,
+           'changes': {},
+           'result': False,
+           'comment': ''}
+
+    if __salt__['saptune.apply_solution'](solution_name=name):
```
arbulu89: We cannot do this. `saptune.apply_solution` will apply the solution, right?
Everything that comes before the `if __opts__['test']:` statement cannot perform any change.

Here, what we should do is have a new method in the modules file like, `is_solution_applied` that returns `True/False`

salt/states/saptune.py
angelabriel: I would prefer 'solution_to_apply' to make it a little bit clearer what we will achieve here.
The term 'solution_is_applied' or even 'is_solution_applied' I would reserve for the function which will check, if a requested solution is already applied or not.

angelabriel:
> 
> And some optional:
> `get_solution_list`
> 

Only a heads up - 'saptune solution list' will give you the list of all available solutions and their related SAP Notes. But if one solution is already applied, this line is highlighted in 'green' color. So you will have some control characters (like ^[[32m) in the output of the command.

Machine readable output of all saptune actions is on our wish list, but currently under discussion and not yet implemented.

salt/states/saptune.py
```diff
@@ -0,0 +1,85 @@
...
+           'changes': {},
+           'result': False,
+           'comment': ''}
+
+    if __salt__['saptune.revert_solution']():
+        ret['result'] = True
+        ret['comment'] = 'Saptune solution is already reverted'
+        return ret
+
+    if __opts__['test']:
```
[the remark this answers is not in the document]
MalloZup: yop I don't get this comment somehow :grin: (to much time ) :rabbit:

salt/modules/saptune.py
MalloZup: I don't get this. I'm using {0} and so on because of the `python` format function. It is like we do here:
https://github.com/SUSE/salt-shaptools/blob/cad04c15571eabdf559045c92fe3044fac84f29d/salt/states/netweavermod.py#L137

salt/states/saptune.py
MalloZup: ok sounds good thx

salt/states/saptune.py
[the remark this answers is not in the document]
MalloZup: ok fixedthis meanwhile

salt/states/saptune.py
```diff
@@ -0,0 +1,85 @@
...
+        ret['comment'] = 'Saptune {} solution would be removed'.format(name)
+        ret['changes']['name'] = name
+        return ret
+
+    try:
+        #  Here starts the actual process
+        result = __salt__['saptune.revert_solution'](
+            solution_name=name,
+
+        if result:
```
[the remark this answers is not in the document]
MalloZup: imho in sake of minimalism we don't need to recheck again since to check if the solution is applied we should trust the CLI saptune

salt/modules/saptune.py
MalloZup: done

salt/modules/saptune.py
```diff
@@ -0,0 +1,71 @@
...
+            False,
+            'The saptune execution module failed to load: the saptune package'
+            ' is not available.')
+
+def is_solution_applied(solution_name):
+    '''
+    check if the saptune solution is applied or not
+    '''
+
+    SOLUTION_TO_SEARCH = "TUNE_FOR_SOLUTIONS=\"{}\"".format(solution_name)
```
arbulu89: Why this variable with capital letters? It's not a constant

salt/modules/saptune.py
```diff
@@ -0,0 +1,71 @@
...
+    if bool(salt.utils.path.which(SAPTUNE_BIN)):
+        return __virtualname__
+
+    else:
+        return (
+            False,
+            'The saptune execution module failed to load: the saptune package'
+            ' is not available.')
+
+def is_solution_applied(solution_name):
```
arbulu89: I don't really know if this is the best way to check if a solution is applied or not. I think @angelabriel can give us a hint if we should use this file or the saptune command to get the applied solutions.

salt/modules/saptune.py
```diff
@@ -0,0 +1,71 @@
...
+
+__virtualname__ = 'saptune'
+
+SAPTUNE_BIN = '/usr/sbin/saptune'
+
+LOGGER = logging.getLogger(__name__)
+
+def __virtual__():
+    '''
+    Only load this module if crm package is installed
```
arbulu89: `s/crm/saptune/`

salt/modules/saptune.py
```diff
@@ -0,0 +1,71 @@
...
+    '''
+    if bool(salt.utils.path.which(SAPTUNE_BIN)):
```
arbulu89: I don't know if we should check if the installed saptune is using version 2 (I guess we won't support the older saptune)

salt/modules/saptune.py
```diff
@@ -0,0 +1,71 @@
...
+
+def is_solution_applied(solution_name):
+    '''
+    check if the saptune solution is applied or not
+    '''
+
+    SOLUTION_TO_SEARCH = "TUNE_FOR_SOLUTIONS=\"{}\"".format(solution_name)
+
+    # open the config file and search if the solution is enabled
+    with open("/etc/syconfig/saptune") as conf:
```
arbulu89: If we use the file, I would put that in a constant as the `SAPTUNE_BIN`

salt/modules/saptune.py
MalloZup: ruby way of things.. :P let me convert to python :)

salt/modules/saptune.py
MalloZup: afaik I remember she suggested it but yop I can change.

salt/modules/saptune.py
MalloZup: @angelabriel cc

salt/modules/saptune.py
MalloZup: make sense

salt/modules/saptune.py
angelabriel: Yes, from my point of view this makes sense.
You can use the command 'saptune version', which will give you the output "current active saptune version is '2'"

Or check variable SAPTUNE_VERSION= in /etc/sysconfig/saptune

salt/modules/saptune.py
angelabriel: yes, at the moment this is the best way. When we will have a specific 'action' or machine readable output, then we can use a better way. But for now we need to stay with this.

angelabriel:
> I have a new concern. What are we going to do if one solution is already applied?
> I would assume that the current solution is reverted and the new one applied (as only 1 solution can exist). This would make the whole logic more complex (except saptune binary is able to manage that, what I don't know).
> I would vote for having this logic (to assure idempotency). What do you think @MalloZup and @angelabriel ?

As we are in a deployment situation, I agree, that this will be the way to handle this situation.
But in a productive SAP environment (so outside of a 'deployment') I would never recommend to automatically revert an already applied solution and apply another one, as you never know which performance impact this will have. Only an administrator should do that (and hopefully he will know what he is doing)

salt/modules/saptune.py
MalloZup: ok thx!

MalloZup:
@angelabriel @arbulu89  Regarding the situation of already applied solution, 

I think the best thing to implement is:

* if there is already a solution applied the call of the module function should error out sayng that there is already a solution applied. In that case the admin should revert the solution before and apply another one

arbulu89:
> @angelabriel @arbulu89 Regarding the situation of already applied solution,
> 
> I think the best thing to implement is:
> 
>     * if there is already a solution applied the call of the module function should error out sayng that there is already a solution applied. In that case the admin should revert the solution before and apply another one

I think currently is the best solution, in the future I would add a new param like `force` to revert the current solution and apply the new, this is a really common scenario in salt

salt/modules/saptunemod.py
```diff
@@ -0,0 +1,80 @@
...
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' saptune.apply_solution solution-name
+    '''
+    cmd = '{} solution apply {}'.format(SAPTUNE_BIN, solution_name)
+    retcode = __salt__['cmd.retcode'](cmd)
+
+    if is_solution_applied(solution_name) and retcode == 0:
```
arbulu89: I would remove this `if/else` statements. The module must only apply changes, not check if it's applied or not, this is part of the state implementation.

salt/modules/saptunemod.py
```diff
@@ -0,0 +1,76 @@
...
+Module to provide Saptune functionality to Salt
+
+.. versionadded:: pending
+
+:maintainer:    Dario Maiocchi <dmaiocchi@suse.com>
+:maturity:      alpha
+:depends:       ``saptune`` Python module
+:platform:      all
+
+:configuration: This module requires the saptune python module and uses the
```
arbulu89: I would just keep: `:configuration: This module requires the saptune package`

salt/modules/saptunemod.py
```diff
@@ -0,0 +1,76 @@
...
+            MINIMAL_SAPTUNE_SUP_VERSION) >= 0
+
+        if use_saptune_supported:
+          return __virtualname__
+
+    else:
+        return (
+            False,
+            'The saptune execution module failed to load: the saptune package'
+            ' is not available.')
```
arbulu89: Maybe add here the `'or the version is older than {}'.format(MINIMAL_SAPTUNE_SUP_VERSION)`

salt/states/saptunemod.py
```diff
@@ -0,0 +1,78 @@
...
+            ret['changes']['name'] = name
+            ret['comment'] = 'Error appling saptune solution'
+            ret['result'] = False
+            return ret
+
+        # check if the solution was applied or not
+
+        if not __salt__['saptune.is_solution_applied'](solution_name=name):
+          ret['result'] = False
+          ret['comment'] = 'Saptune solution was not applied correctly. Perhaps an already solution need to be reverted first'
```
arbulu89: `Perhaps an already applied solution...`

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def setup_loader_modules(self):
+        return {saptune: {}}
+
+    @mock.patch('salt.utils.path.which')
+    def test_virtual_saptune(self, mock_which):
+        mock_pkg_version = MagicMock(return_value='2.0.0')
+        mock_pkg_version_cmp = MagicMock(return_value=2)
+
+        mock_which.side_effect = [True, True]
```
arbulu89: Why this `side_effect`? As far as I see the `which` method is used only once. You could put here `mock_which.return_value = True` what would make the same.

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,85 @@
...
+        with patch.dict(saptune.__salt__, {
+                'pkg.version': mock_pkg_version,
+                'pkg.version_cmp': mock_pkg_version_cmp}):
+            assert saptune.__virtual__() == 'saptune'
+            mock_which.assert_called_once_with(saptune.SAPTUNE_BIN)
+
+    @mock.patch('salt.utils.path.which')
+    def test_virtual_saptune_error(self, mock_which):
```
arbulu89: This test doesn't make much sense. You don't really need to have so many code. just with the next would be enough:

```
mock_which.return_value = False
assert saptune.__virtual__() == (
    False, 'The saptune execution module failed to load: the saptune package'
     ' is not available.')
mock_which.assert_called_once_with(saptune.SAPTUNE_BIN)
```

salt/modules/saptunemod.py
```diff
@@ -0,0 +1,76 @@
...
+    '''
+    Only load this module if saptune package is installed
+    '''
+    if bool(salt.utils.path.which(SAPTUNE_BIN)):
+        version = __salt__['pkg.version']('saptune')
+        # don't support older then 2.0
+        use_saptune_supported = __salt__['pkg.version_cmp'](version,
+            MINIMAL_SAPTUNE_SUP_VERSION) >= 0
+
+        if use_saptune_supported:
```
arbulu89: What happens here if `use_saptune_supported` is `False`? It wouldn't return anything, right?
If this is the case we need to return the same as the other error.

Just removing the next `else` statement would be enough

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,85 @@
...
+            assert response == (
+              False, 'The saptune execution module failed to load: the saptune package'
+              ' is not available.')
+            mock_which.assert_called_once_with(saptune.SAPTUNE_BIN)
+
+    def test_is_solution_applied_return_false(self):
+        '''
+        Test is_solution_applied method return false
+        '''
+        saptune.is_solution_applied = MagicMock(return_value=False)
```
arbulu89: This UT is wrong. You cannot mock what you are testing hehe
This UT will pass even though you change the code under test.
You will need to mock the `open` and `read` methods

tests/unit/modules/test_saptunemod.py
MalloZup: yop.I followed the same pattern as: https://github.com/SUSE/salt-shaptools/blob/master/tests/unit/modules/test_crmshmod.py#L50

MalloZup: i did as suggest in comment thx

MalloZup:
regarding the UT, i already tried to look at the existing codebase but personally I think that there is no clear emerging pattern for a nib among the codebase i felt we have diverging pattern even in the unit-test (here).
I will try re-look and perhaps improve doc if I can. thx for pointers!

MalloZup:
@arbulu89 I think is good to merge right now. I have update the test etc. let me know! tia! :rocket:

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,95 @@
...
+        return {saptune: {}}
+
+    def tearDown(self):
+        try:
+            os.remove(FAKE_SAPTUNE_CONF)
+        except:
+            pass
+
+    @mock.patch('salt.utils.path.which')
+    def test_virtual_saptune(self, mock_which):
```
arbulu89: I think that the use case where `saptune` exists but the version is lower than 2 is not covered

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,95 @@
...
+        mock_which.return_value = False
+        assert saptune.__virtual__() == (
+            False, 'The saptune execution module failed to load: the saptune package'
+             ' is not available, or the version is older than 2.0.0')
+    
+    def test_is_solution_applied_return_false(self):
+        '''
+        Test is_solution_applied method return false
+        '''
+        saptune.SAPTUNE_CONF = FAKE_SAPTUNE_CONF
```
arbulu89: This is a really bad practice. You are changing the value of one constant permanently here, so the next tests might be affected somehow (maybe not in this particular file).

Besides that, I'm not a big fan of using this kind of techniques (replace files) to unit testing. In any case, we should use real example files to test, and in the best scenario mock the `open` method.

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,95 @@
...
+        with open(FAKE_SAPTUNE_CONF, 'w') as conf:  
+            conf.write("TUNE_FOR_SOLUTIONS=\"{}\"".format(solution_name))
+        assert saptune.is_solution_applied(solution_name)
+
+    def test_apply_solution(self):
+        '''
+        Test apply solution method 
+        '''
+        
+        mock_cmd = MagicMock(return_value=True)
```
arbulu89: `retcode` would return a number instead of a boolean, but for the UT it doesn't really affect

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+    def test_solution_applied_true(self):
+        '''
+        Test solution is applied method true
+        '''
+
+        response = MagicMock(return_value=True)
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': response}):
+            assert saptune.solution_applied('fakesolution')
```
arbulu89: This `assert` is not correct. Here you need to test the returned dictionary. Salt states method always return a dictionary, and we need to test the result, comment, etc. This `assert` will be true always otherwise (except Exception raised what shouldn't happen ofter)

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,55 @@
...
+        '''
+        Test solution is applied method false
+        '''
+
+        response = MagicMock(return_value=False)
+        solution_name = 'testmode'
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': response}):
+            with patch.dict(saptune.__opts__, {'test': True}):
+              ret = saptune.solution_applied(solution_name)
+              assert ret['comment'] == "Saptune {} solution would be applied".format(solution_name)
```
arbulu89: Pleas, assert the other values of the dictionary too

README.md
```diff
@@ -33,12 +33,15 @@ To run the module funcionalities in the minions:
 2. Copy the content of **states** in your "salt://_states/" (by default /srv/salt/_states)
 3. Synchronize modules with the minions. For that run:
 ```bash
-sudo salt '*' saltutil.sync_states
+sudo salt-call  saltutil.sync_all
```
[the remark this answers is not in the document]
MalloZup: without this changes we can't load the module we load just the state.

salt/modules/saptunemod.py
```diff
@@ -0,0 +1,74 @@
...
+          return __virtualname__
+
+    return (
+            False,
+            'The saptune execution module failed to load: the saptune package'
+            ' is not available, or the version is older than {}'.format(MINIMAL_SAPTUNE_SUP_VERSION))
+
+def is_solution_applied(solution_name):
+    '''
+    check if the saptune solution is applied or not
```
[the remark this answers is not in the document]
MalloZup: done

tests/unit/modules/test_saptunemod.py
MalloZup: https://github.com/SUSE/salt-shaptools/pull/45/files#diff-1630922ee48bdbca8646c751099813fcR58

tests/unit/states/test_saptunemod.py
MalloZup: done

tests/unit/states/test_saptunemod.py
MalloZup: done thx

README.md
```diff
@@ -71,6 +86,21 @@ sudo chmod 755 tests/run.sh
...
+For testing modules:
+
+```
+salt-call --local saptune.apply_solution "SAP-ASE"
+
+```
+
+For testing states:
+```
+salt-call --local state.single saptune.solution_applied "HANa"
```
arbulu89: `s/HANa/HANA` maybe better

README.md
```diff
@@ -71,6 +86,21 @@ sudo chmod 755 tests/run.sh
 ./tests/run.sh
 ```
 
+4). Testing manually:
```
arbulu89: I wouldn't call `Testing manually`, I would `Running your modules/states`

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,124 @@
...
+            conf.write("NO_SOLUTION")
+            # we don't find the solution foo in file so it should be false
+        assert not saptune.is_solution_applied('foo')
+
+    def test_is_solution_applied_return_true(self):
+        '''
+        Test is_solution_applied method return true
+        '''
+        solution_name = 'ok_solution'
+        saptune.SAPTUNE_CONF = FAKE_SAPTUNE_CONF
```
arbulu89: Add an extra space here. This issue are usually highlighted if you have a proper linter installed in you code editor (pylint, pep8)

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,124 @@
...
+        '''
+        Test is_solution_applied method return false
+        '''
+        saptune.SAPTUNE_CONF = FAKE_SAPTUNE_CONF
+        with open(FAKE_SAPTUNE_CONF, 'w') as conf: 
+            conf.write("NO_SOLUTION")
+            # we don't find the solution foo in file so it should be false
+        assert not saptune.is_solution_applied('foo')
+
+    def test_is_solution_applied_return_true(self):
```
arbulu89: This test is doing nothing, right?

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,124 @@
...
+        '''
+        Test is_solution_applied method return true
+        '''
+        solution_name = 'ok_solution'
+        saptune.SAPTUNE_CONF = FAKE_SAPTUNE_CONF
+    def test_is_solution_applied_return_false(self):
```
arbulu89: Is this test repeated?

tests/unit/modules/test_saptunemod.py
```diff
@@ -0,0 +1,124 @@
...
+    def test_apply_solution(self):
+        '''
+        Test apply solution method 
+        '''
+        
+        mock_cmd = MagicMock(return_value=True)
+        with patch.dict(saptune.__salt__, {'cmd.retcode': mock_cmd}):
+            assert saptune.apply_solution('foo')
+
+    def test_apply_solution_false(self):
```
arbulu89: This test doesn't make sense. `apply_solution` will only return an integer (0 if correct, other number, otherwise). I would only have one test returning an integer. The tests must follow the sense of the code

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,126 @@
...
+          'comment': "Saptune solution applied"}
+          
+        sol_applied_resp = MagicMock(return_value=True)
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': sol_applied_resp}):
+            assert saptune.solution_applied(solution_name) == expected_ret
+
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': MagicMock(side_effect=[False, True]), 
+                                            'saptune.apply_solution': MagicMock(return_value=0)}):
+                assert saptune.solution_applied(solution_name) == expected_ret2
+
```
arbulu89: Why so many new lines? Only 1 line should go here to follow the code quality guidelines

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,126 @@
...
+          'name': solution_name,
+          'comment': "Saptune solution applied"}
+          
+        sol_applied_resp = MagicMock(return_value=True)
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': sol_applied_resp}):
+            assert saptune.solution_applied(solution_name) == expected_ret
+
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': MagicMock(side_effect=[False, True]), 
+                                            'saptune.apply_solution': MagicMock(return_value=0)}):
+                assert saptune.solution_applied(solution_name) == expected_ret2
```
arbulu89: I think the indentation is not good here. 4 spaces should go in this assert line

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,126 @@
...
+        Test solution is applied method already
+        '''
+        solution_name = 'normalmode'
+
+        expected_ret = { 'changes': {}, 
+          'result': False,
+          'name': solution_name,
+          'comment': 'Saptune solution was not applied correctly. Perhaps an already applied solution need to be reverted first'}
+
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': MagicMock(return_value=0),
```
arbulu89: What are you exactly testing here?
I think you are testing the scenario where the 2nd `is_solution_applied` returns False, right?
For that case, update the return_value of `is_solution_applied` to side_effect=[False, False]
Remember that `is_solution_applied` returns a boolean, not a number

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,126 @@
...
+
+        expected_ret = { 'changes': {}, 
+          'result': False,
+          'name': solution_name,
+          'comment': 'Saptune solution was not applied correctly. Perhaps an already applied solution need to be reverted first'}
+
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': MagicMock(return_value=0), 
+                                           'saptune.apply_solution': MagicMock(return_value=0)}):
+                assert saptune.solution_applied(solution_name) == expected_ret
+
```
arbulu89: Only 1 new line

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,126 @@
...
+        solution_name = 'normalmode'
+        expected_ret = { 'changes': {}, 
+          'result': False,
+          'name': solution_name,
+          'comment': "Error appling saptune solution"} 
+    
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': MagicMock(return_value=False), 
+                                            'saptune.apply_solution': MagicMock(return_value=1)}):
+            assert saptune.solution_applied(solution_name) == expected_ret
+
```
arbulu89: Only 1 new line

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,126 @@
...
+               'name':  solution_name,
+               'comment': "Saptune {} solution would be applied".format(solution_name)}
+
+        response = MagicMock(return_value=False)
+     
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied': response}):
+            with patch.dict(saptune.__opts__, {'test': True}):
+              ret = saptune.solution_applied(solution_name)
+              assert ret == expected_ret
+
```
arbulu89: Only 1 new line

tests/unit/states/test_saptunemod.py
```diff
@@ -0,0 +1,126 @@
...
+        Test solution is applied method error exception 
+        '''
+        solution_name = 'normalmode'
+
+        expected_ret = { 'changes': {}, 
+          'result': False,
+          'name': solution_name,
+          'comment': 'saptune error'}
+
+        with patch.dict(saptune.__salt__, {'saptune.is_solution_applied':  MagicMock(side_effect=[False, 1]),
```
arbulu89: Why `side_effect=[False, 1]`, all the return codes must be boolean

tests/unit/modules/test_saptunemod.py
arbulu89: This is still pending to be addressed

MalloZup:
thx!  i will merge after done with manual test

MalloZup:
tested worked fine

PR #46, opened 2019-12-19
Add fn to install hana with xml passwords file

Simranpal:
Module and state functionality to install HANA with passwords from XML file if provided
Supports:
- External/already existing xml file
- Generating a new xml file which stores passwords if no existing file provided

Todo: Add option to remove the XML file


Simranpal:
Tested with deployment

salt/states/hanamod.py
```diff
@@ -177,11 +180,18 @@ def installed(
         to the minion and used in the installation. Values in this file might
         be changed setting then in the extra_parameters dictionary using the
         exact name as in the config file as a key
+    hdb_pwd_file
+        If hdb_pwd_file paremeter is set, it will be downloaded from the master
+        to the minion and used in the installation. Values in this file might
+        be changed setting then in the extra_parameters dictionary using the
+        exact name as in the XML file as a key
+    remove_pwd_files
+        If remove the XML password hdb_pwd_file after installation
```
arbulu89: Maybe add at the end that by default the value is `True`

PR #48, opened 2019-12-20
DRBD: check unfinished request of sync status

nick-wang:
The sync target change status to UpToDate first
before syncing md. It may take unexpected long time
on lower level devices.
Checking the unfinished request even state is UpToDate.

Be aware still have possibility to make inconsistent
state of sync target and sync source. But this will
distinct drop the reason cause by slow backing device.

### DIFF
--- a/salt/modules/drbdmod.py
+++ b/salt/modules/drbdmod.py
@@ -278,6 +278,33 @@ def _is_peers_uptodated(name, peernode='all'):
     return ret
 
 
+def _is_no_backing_dev_request(name):
+    '''
+    Check whether all volumes have no unfinished backing device request.
+    Only working when json status supported.
+
+    Metadata still need to sync to disk after state changed.
+    Only reply to sync target to change when I/O request finished,
+    which is unpredictable. Local refernece count is not 0 before endio.
+
+    '''
+    if not __salt__['drbd.json']:
+        return True
+
+    output = OUTPUT_OPTIONS['json']
+    res = output["get_res_func"](name)
+
+    if not res:
+        return False
+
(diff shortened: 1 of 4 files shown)


salt/modules/drbdmod.py
```diff
@@ -690,7 +718,7 @@ def check_sync_status(name, peernode='all'):
         salt '*' drbd.check_sync_status <resource name> <peernode name>
     '''
     if _is_local_all_uptodated(name) and _is_peers_uptodated(
```
arbulu89: Just a question here. Are you running the `drbdsetup status` in all of the methods? For example, here you would run the command three time in a sequence.
Maybe we can avoid that, just passing the output of the command to these methods. This would reduce the executions, and maybe avoid some strange situation where, the status changes between calls (not really probable, but it might happen)

nick-wang: Indeed, it is a waste to run 3 times, since we could use the same result.

nick-wang:
> Thanks @nick-wang -wang ,
> More than the logic itself, that looks good (I don't really know how to interpret the drbdsetup outputs), I think we are running to many times the `status` command. In each `if` condition more than once.
> I would suggest to change this somehow, just getting the status first, and passing this status to the check commands instead of running the command every time.
> What do you think?
> If it's a big change, we can merge this PR and work on that later (if you think that it's a good idea)
Yes, it is a good idea. It doesn't need to call 3 times.
I will fix it in the next week, most effort is adapt the unit test and real env testing i think. So we can either release the pr first(already tested) or wait for the new commits together.

Issue #49, opened 2019-12-20
drbdsetup status command is called multiple times with the same purpose


arbulu89:
The drbd modules run the command `drbdsetup status` to get the current status and operate according this. Nothing wrong here.
The issue is that the command is executed in many methods that are used almost in the same moment. For example:

https://github.com/SUSE/salt-shaptools/blob/master/salt/modules/drbdmod.py#L692

This is not so optimal. A possible option would be to just pass the output of the command to this methods, to use the same data, and avoid multiple executions (and possible changes during the executions creating strange scenarios).

TODO:
Find all the method where `drbdsetup status` (or any other method used to retrieve the status) and update the code to use the result data instead of running the method

PR #50, opened 2019-12-23
Issue49 call drbdsetup once

nick-wang:
Issue https://github.com/SUSE/salt-shaptools/issues/49, avoid multiple executions of drbdsetup.

### DIFF
--- a/salt/modules/drbdmod.py
+++ b/salt/modules/drbdmod.py
@@ -223,48 +223,29 @@ def _line_parser(line):
     func(line)
 
 
-def _is_local_all_uptodated(name):
+def _is_local_all_uptodated(res, output):
     '''
-    Check whether all local volumes are UpToDate.
+    Check whether all local volumes of given resource are UpToDate.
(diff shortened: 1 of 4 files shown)


salt/modules/drbdmod.py
```diff
@@ -717,8 +692,22 @@ def check_sync_status(name, peernode='all'):
 
         salt '*' drbd.check_sync_status <resource name> <peernode name>
     '''
-    if _is_local_all_uptodated(name) and _is_peers_uptodated(
-            name, peernode=peernode) and _is_no_backing_dev_request(name):
+    if __salt__['drbd.json']:
```
arbulu89: Maybe I would put this piece of code in other method like `get_status` (from this `if` to `res = resources[0]`), but this is good too

nick-wang: I was thinking about this, but i also need to get the output[]. Either in this function again or return from _get_status is not good.

PR #58, opened 2020-03-20
Feature/extract pydbapi

arbulu89:
Some new salt methods to extract the python dbapi SAP package.
The method automatically finds the location of the package using the LABELIDX and LABEL files.

The `reload_hdb_connector` is a helper method to reload that library. Otherwise, during salt installation the recently installed py dbapi libraries are not available. It's a small trick to make everything work in the same python process (salt process basically)

Related to:
https://github.com/SUSE/saphanabootstrap-formula/pull/75
https://github.com/SUSE/sapnwbootstrap-formula/pull/42

### DIFF
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -27,9 +27,9 @@ import logging
 import time
 import re
 
-try:
+try:  # pragma: no cover
     import importlib as imp
-except ImportError:
+except ImportError:  # pragma: no cover
     import imp
 
 from salt import exceptions
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -945,7 +945,7 @@ def _find_sap_folder(software_folders, folder_pattern):
                     return folder
                 else:
                     LOGGER.debug('%s folder does not containt HANA client', folder)
-        except FileNotFoundError:
+        except IOError:
             LOGGER.debug('%s file not found in %s. Skipping folder', LABEL_FILE, folder)
 
         labelidx = '{}/{}'.format(folder, LABELIDX_FILE)
@@ -958,7 +958,7 @@ def _find_sap_folder(software_folders, folder_pattern):
                     return _find_sap_folder(new_folders, folder_pattern)
                 except HanaClientNotFound:
                     continue
-        except FileNotFoundError:
+        except IOError:
             LOGGER.debug('%s file not found in %s. Skipping folder', LABELIDX_FILE, folder)
 
     raise HanaClientNotFound('HANA client not found')
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -26,6 +26,7 @@ from __future__ import absolute_import, unicode_literals, print_function
 import logging
 import time
 import re
+import imp
 
 from salt import exceptions
 from salt.utils import files as salt_files
@@ -917,6 +918,14 @@ def wait_for_connection(
             'HANA database not available after {} seconds in {}:{}'.format(
                 timeout, host, port
             ))
+
+
+def reload_hdb_connector():
+    '''
+    As hdb_connector uses pyhdb or dbapi, if these packages are installed on the fly,
+    we need to reload the connector to import the correct api
+    '''
+    imp.reload(hdb_connector)
 
 
 def _find_sap_folder(software_folders, hana_client_pattern):
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -26,6 +26,7 @@ from __future__ import absolute_import, unicode_literals, print_function
 import logging
 import time
 import re
+
 import imp
 
 from salt import exceptions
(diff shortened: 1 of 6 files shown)


salt/modules/hanamod.py
```diff
@@ -900,4 +921,71 @@ def wait_for_connection(
...
+    '''
+    for folder in software_folders:
+        label = '{}/{}'.format(folder, LABEL_FILE)
+        try:
+            with salt_files.fopen(label, 'r') as label_file_ptr:
+                label_content = label_file_ptr.read().strip()
+                if folder_pattern.match(label_content):
+                    return folder
+                else:
+                    LOGGER.debug('%s folder does not containt HANA client', folder)
```
Simranpal: Maybe this can be more generic as we are not searching HANA client always

salt/modules/hanamod.py
```diff
@@ -900,4 +921,71 @@ def wait_for_connection(
...
+        software_folders,
+        output_dir,
+        hana_version='20'):
+    '''
+    Extract HANA pydbapi python client from the provided software folders
+
+    name
+        Name of the package that needs to be installed
+    software_folders
+        Folders where the
```
Simranpal: The description looks to be incomplete

salt/modules/hanamod.py
```diff
@@ -900,4 +921,71 @@ def wait_for_connection(
...
+    '''
+    current_platform = hana.HanaInstance.get_platform()
+    hana_client_pattern = re.compile('^HDB_CLIENT:{}.*:{}:.*'.format(
+        hana_version, current_platform))
+    try:
+        hana_client_folder = _find_sap_folder(software_folders, hana_client_pattern)
+    except HanaClientNotFound:
+        raise exceptions.CommandExecutionError('HANA client not found')
+    pydbapi_file = '{}/client/{}'.format(hana_client_folder, name)
+    __salt__['archive.tar'](options='xvf', tarfile=pydbapi_file, dest=output_dir)
```
Simranpal: Is is ok to hardcode the `archive.tar` filename

salt/states/hanamod.py
```diff
@@ -755,3 +755,49 @@ def memory_resources_updated(
...
+        output_dir,
+        hana_version='20',
+        force=False):
+    '''
+    Extract HANA pydbapi python client from the provided software folders
+
+    name
+        Name of the package that needs to be installed
+    software_folders
+        Folders where the
```
Simranpal: The description looks incomplete

salt/states/hanamod.py
```diff
@@ -755,3 +755,49 @@ def memory_resources_updated(
     except exceptions.CommandExecutionError as err:
         ret['comment'] = six.text_type(err)
         return ret
+
+
+def pydbapi_extracted(
+        name,
+        software_folders,
+        output_dir,
+        hana_version='20',
```
Simranpal: Should version not start `2xxx`. Actually I see its` HDB_CLIENT:20xxx` from LABEL

salt/modules/hanamod.py
```diff
@@ -900,4 +921,71 @@ def wait_for_connection(
...
+
+        labelidx = '{}/{}'.format(folder, LABELIDX_FILE)
+        try:
+            with salt_files.fopen(labelidx, 'r') as labelidx_file_ptr:
+                labelidx_content = labelidx_file_ptr.read().splitlines()
+                new_folders = [
+                    '{}/{}'.format(folder, new_folder) for new_folder in labelidx_content]
+                try:
+                    return _find_sap_folder(new_folders, folder_pattern)
+                except HanaClientNotFound:
```
diegoakechi: Same here. Maybe something like HANA components not found + the pattern that was being searched. Also I think it can be reused on the implementation to find the hdblcm binary, can't it?

salt/modules/hanamod.py
diegoakechi: I think `archive.tar` here is the salt execution module being called. The file name is passed as parameter, so I think it is correct.

BTW, the sapcar feature should follow similar approach to be consistent with other compression utilities already existent on salt.

salt/modules/hanamod.py
arbulu89: Yes, I changed some other variables but not the error message

salt/modules/hanamod.py
arbulu89: Yes, `archive.tar` is the used method. There is no other name hehe

salt/modules/hanamod.py
arbulu89: Mmm, I was thinking about that. Anyway, this will come in a future PR as it's other feature implementation. If we see it might be reused (I'm not sure) I will move it even to `shaptool` project

salt/modules/hanamod.py
arbulu89: @diegoakechi @Simranpal In fact, it might be a good idea to call the module `archive.sapcar`. We will discuss this in the implementation PR itself

arbulu89:
@diegoakechi @Simranpal I have done the changes suggested by you.
About some open endings.
1. The `find_sap_folder` method might be used to find the HANA method, but it would require some drastic changes there (unfortunately, the HANA software doesn't have LABEL and LABELIDX files). Most probably I will use other strategy.
2. If the methods can be used elsewhere. Of course. I have not found any other use case by now, so it was easy to put it in the hana code, but we can definetely find other location. But I would do it in the future if we need so in any case.

PR #59, opened 2020-03-22
Modules for sapcar extraction

Simranpal:
Add the salt functions to extract the SAP sar package. 

- I created new file names as `saputils` to add future sap related functions. The option to extract rar,exe SAP installation package can also be added here, what do you think?

- And I kept the `output_dir` is optional, and if its not provided, then use current directory as sar_file to extract.

TO-Do:
Add unit tests and make any further improvements

Related to:
https://github.com/SUSE/sapnwbootstrap-formula/pull/41
https://github.com/SUSE/shaptools/pull/49


salt/modules/saputilsmod.py
```diff
@@ -0,0 +1,64 @@
...
+    '''
+    if HAS_SAPUTILS:
+        return __virtualname__
+    return (
+        False,
+        'The saputils execution module failed to load: the shaptools python'
+        ' library is not available.')
+
+
+def extract_sapcar_file(
```
arbulu89: The functionality looks fine. I would rename the method directly to `extract` or `sapcar` following the convention here:
https://github.com/saltstack/salt/blob/master/salt/modules/archive.py#L466

`sapcar.sapcar` would be a bit strange, that's true

salt/modules/saputilsmod.py
```diff
@@ -0,0 +1,64 @@
...
+from salt import exceptions
+
+# Import third party libs
+try:
+    from shaptools import saputils
+    HAS_SAPUTILS = True
+except ImportError:  # pragma: no cover
+    HAS_SAPUTILS = False
+
+__virtualname__ = 'saputils'
```
arbulu89: Rename to `sapcar`

salt/states/saputilsmod.py
```diff
@@ -0,0 +1,102 @@
...
+from salt import exceptions
+from salt.ext import six
+
+try:
+    from shaptools import saputils
+    HAS_SAPUTILS = True
+except ImportError:  # pragma: no cover
+    HAS_SAPUTILS = False
+
+__virtualname__ = 'saputils'
```
arbulu89: Rename to `sapcar`

salt/states/saputilsmod.py
```diff
@@ -0,0 +1,102 @@
...
+    Only load this module if shaptools python module is installed
+    '''
+    if HAS_SAPUTILS:
+        return __virtualname__
+    return (
+        False,
+        'The saputils execution module failed to load: the shaptools python'
+        ' library is not available.')
+
+def sar_file_extracted(
```
arbulu89: Rename to `extracted`

salt/states/saputilsmod.py
```diff
@@ -0,0 +1,102 @@
...
+        name,
+        sapcar_exe,
+        sar_file,
+        output_dir=None,
+        options=None):
+    """
+    Extract a SAPCAR sar archive
+
+    name
+        SAR file name to ber extracted
```
arbulu89: `s/ber/be/`

salt/states/saputilsmod.py
```diff
@@ -0,0 +1,102 @@
...
+    sapcar_exe
+        Path to the SAPCAR executable file. SAPCAR is a SAP tool to extract SAP SAR format archives 
+    sar_file
```
arbulu89: What is the usage of this `sar_file`?
Use just the `name`, right?

salt/states/saputilsmod.py
```diff
@@ -0,0 +1,102 @@
...
+    
+    try:
+        #  Here starts the actual process
+        __salt__['saputils.extract_sapcar_file'](
+            sapcar_exe=sapcar_exe,
+            sar_file=sar_file,
+            output_dir=output_dir,
+            options=options)
+        if output_dir:
+            ret['changes']['output_dir'] = output_dir
```
arbulu89: You could just have `ret['changes']['output_dir'] = output_dir or os.path.dirname(sar_file)` or some like that. Current solution is good though

salt/modules/saputilsmod.py
Simranpal: Thanks @arbulu89 `sar` is another option, but I will rename to `extract` for better understanding

salt/states/saputilsmod.py
Simranpal: yes, I changed to just use `name`. Does it make sense to also rename `sar_file` parameter to `name` in salt module fn and shaptools to keep it consistent?

salt/states/sapcarmod.py
```diff
@@ -0,0 +1,97 @@
...
+try:
+    from shaptools import saputils
+    HAS_SAPUTILS = True
+except ImportError:  # pragma: no cover
+    HAS_SAPUTILS = False
+
+__virtualname__ = 'sapcar'
+
+
+def __virtual__():  # pragma: no cover
```
arbulu89: The virtual method is not correct. You shouldn't check for `shaptools`. You have to check if the `sapcar` salt module is available.
Check in https://github.com/SUSE/salt-shaptools/blob/master/salt/states/crmshmod.py#L45

arbulu89:
@diegoakechi We are not following the `archive.tar` conventions here. I'm quite happy with this naming and methods. Otherwise, we would need change the module to have something like `sapcar` instead of `extract` and pass the options from the state. We can do that aswell.
What do you think @Simranpal @diegoakechi ?
https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.archive.html
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.archive.html
PD: The state naming is the same

diegoakechi:
@arbulu89 It was more a suggestion from my side, but I am ok to proceed as it is right now.

salt/states/sapcarmod.py
Simranpal: Thanks @arbulu89 yes I followed the approach previously, but got some error most likely because I was setting incorrect return type. 
However, I have corrected it now.

PR #60, opened 2020-03-31
Fix spec file with sapcar modules

Simranpal:
In the last PR adding sapcar functionality (https://github.com/SUSE/salt-shaptools/pull/59)
I did not update the `install` of package salt-shaptools. With this fix, I add the relevant sapcar modules files.


Simranpal:
Thanks @arbulu89 I tested it now with deployment in libvirt and works fine

PR #61, opened 2020-04-24
Feature/improve pydbapi extraction

Simranpal:
This PR brings two changes to pydbapi extraction functions:
-Add support for tar options
-Raise error when `software_folders` parameter to the functions is when not passed as list type

Also Added the UT for this change and updated the existing UTs

Note that I have added the extra tar options preceeding the `xvf` extract option which requires `xvf tarfile` together

Salt tar functionality:  https://github.com/saltstack/salt/blob/779bc179e25be49d3449a56648bf827bcb6d5fb6/salt/modules/archive.py#L466

Todo: update spec file, change log

### DIFF
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -970,6 +970,7 @@ def extract_pydbapi(
         name,
         software_folders,
         output_dir,
+        tar_options=None,
         hana_version='20'):
     '''
     Extract HANA pydbapi python client from the provided software folders
@@ -982,14 +983,21 @@ def extract_pydbapi(
         standard way in SAP landscape
     output_dir
         Folder where the package is extracted
+    extract_options
+        Additional options to pass to the tar extraction command
     '''
     current_platform = hana.HanaInstance.get_platform()
+    tar_options_str = '{} xvf'.format(tar_options) if tar_options else 'xvf'
     hana_client_pattern = re.compile('^HDB_CLIENT:{}.*:{}:.*'.format(
         hana_version, current_platform))
+    if not isinstance(software_folders, list):
+        raise TypeError(
+            "software_folders must be list, not {} type".format(type(software_folders).__name__)
+        )
     try:
         hana_client_folder = _find_sap_folder(software_folders, hana_client_pattern)
     except SapFolderNotFoundError:
         raise exceptions.CommandExecutionError('HANA client not found')
     pydbapi_file = '{}/client/{}'.format(hana_client_folder, name)
-    __salt__['archive.tar'](options='xvf', tarfile=pydbapi_file, dest=output_dir)
+    __salt__['archive.tar'](options=tar_options_str, tarfile=pydbapi_file, dest=output_dir)
     return pydbapi_file
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -986,14 +986,14 @@ def extract_pydbapi(
     extract_options
         Additional options to pass to the tar extraction command
     '''
-    current_platform = hana.HanaInstance.get_platform()
-    tar_options_str = '{} xvf'.format(tar_options) if tar_options else 'xvf'
-    hana_client_pattern = re.compile('^HDB_CLIENT:{}.*:{}:.*'.format(
-        hana_version, current_platform))
     if not isinstance(software_folders, list):
         raise TypeError(
             "software_folders must be list, not {} type".format(type(software_folders).__name__)
         )
+    current_platform = hana.HanaInstance.get_platform()
+    tar_options_str = '{} xvf'.format(tar_options) if tar_options else 'xvf'
+    hana_client_pattern = re.compile('^HDB_CLIENT:{}.*:{}:.*'.format(
+        hana_version, current_platform))
     try:
         hana_client_folder = _find_sap_folder(software_folders, hana_client_pattern)
     except SapFolderNotFoundError:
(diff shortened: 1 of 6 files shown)


salt/modules/hanamod.py
```diff
@@ -982,14 +983,21 @@ def extract_pydbapi(
         standard way in SAP landscape
     output_dir
         Folder where the package is extracted
+    extract_options
+        Additional options to pass to the tar extraction command
     '''
+    if not isinstance(software_folders, list):
+        raise TypeError(
+            'software_folders must be list, not {} type'.format(type(software_folders).__name__)
```
arbulu89, line 991: Do we need the `__name__` extension? I wouldn't add it. I think the output of `type` is quite good. Anyway, up to you, this is my personal opinion

salt/modules/hanamod.py
```diff
@@ -982,14 +983,21 @@ def extract_pydbapi(
...
+        )
     current_platform = hana.HanaInstance.get_platform()
+    tar_options_str = '{} xvf'.format(tar_options) if tar_options else 'xvf'
```
arbulu89, line 994: This is confusing. The new parameter is used to add additional options to the `tar` operation. In this case, the name should be something else i think, `additional_extract_options` or something like that.
We need to document the usage much better, we have `default` parameters too.

salt/states/hanamod.py
```diff
@@ -756,11 +756,11 @@ def memory_resources_updated(
         ret['comment'] = six.text_type(err)
         return ret
 
-
 def pydbapi_extracted(
         name,
         software_folders,
         output_dir,
+        tar_options=None,
```
arbulu89, line 763: Put it at the end to make it backward compatible. And the name doesn't look like good. They are additional options, as `xvf` is always used. We need to change something here to make it logical

tests/unit/modules/test_hanamod.py
```diff
@@ -892,13 +892,13 @@ def test_extract_pydbapi(self, mock_get_platform, mock_find_sap_folders, mock_co
         mock_tar = MagicMock()
         with patch.dict(hanamod.__salt__, {'archive.tar': mock_tar}):
             pydbapi_file = hanamod.extract_pydbapi(
-                'PYDBAPI.tar.gz', ['1234', '5678'], '/tmp/output')
+                'PYDBAPI.tar.gz', ['1234', '5678'], '/tmp/output', '-l')
```
arbulu89, line 895: Please check if we should pass `-l` or `l`. Even though it's just a unit test, the more realistic the better.

https://github.com/saltstack/salt/blob/master/salt/modules/archive.py#L494

salt/modules/hanamod.py
Simranpal: yes, the type() returns the class type of the argument. It returns something like `<class 'str'>` otherwise

salt/modules/hanamod.py
Simranpal: yes I will change the parameter to `additional_extract_options`

tests/unit/modules/test_hanamod.py
Simranpal: Based on my test, both options worked with the gnu `tar` and salt `tar` function
However I will change it to `l` to keep consistent with our default tar options

Simranpal: Actually I take that last comment back. On further testing: when passing multiple tar options, it does not work as expected and I was facing tar errors. So in the case to support multiple tar options we need to add the `-` prefixing. 
I updated this and now the options will look like `tar -l -xvf tarfile` instead of `tar l xvf tarfile`

Simranpal:
Hi @arbulu89 I did some further changes to the functions. Let me know what you think

salt/modules/hanamod.py
```diff
@@ -982,14 +983,22 @@ def extract_pydbapi(
         standard way in SAP landscape
     output_dir
         Folder where the package is extracted
+    additional_extract_options
+        Additional options to pass to the tar extraction command
     '''
+    if not isinstance(software_folders, list):
+        raise TypeError(
+            "software_folders must be list, not {} type".format(type(software_folders).__name__)
```
arbulu89, line 991: Better `...must be a list, ...`, right?

salt/modules/hanamod.py
```diff
@@ -982,14 +983,22 @@ def extract_pydbapi(
...
+        )
     current_platform = hana.HanaInstance.get_platform()
+    tar_options_str = ('{} -xvf'.format(additional_extract_options)
```
arbulu89, line 994: Do we need to put this code into parenthesis? This will create a tuple. Is this what we want? We were passing a simple string before

salt/modules/hanamod.py
Simranpal: @arbulu89 you mean remove the `, not {} type` this removes the need to find the type

salt/modules/hanamod.py
Simranpal: Hi @arbulu89 I put this in parathesis , following Python's implied line continuation inside parentheses
https://www.python.org/dev/peps/pep-0008/#maximum-line-length
I checked the type for `tar_options_str` is `str`
Is this not correct

arbulu89: You are absolutely right!
Thanks for checking this out.

salt/modules/hanamod.py
arbulu89: No, only that I miss an `a` between `be` and `list`, to be `must be a list`.

Simranpal:
Also tested with the deployments and changes worked as expected

PR #63, opened 2020-05-21
Feature/configure cluster properties

arbulu89:
Add new states to configure the cluster properties and defaults.

Besides that, I have added the `force` flag to `cluster_configured` to avoid hanging there forever as crmsh asks for user interactions

https://bugzilla.suse.com/show_bug.cgi?id=1170702

### DIFF
--- a/salt/modules/crmshmod.py
+++ b/salt/modules/crmshmod.py
@@ -23,6 +23,7 @@ Module to provide CRM shell (HA cluster) functionality to Salt
 # Import Python libs
 from __future__ import absolute_import, unicode_literals, print_function
 import logging
+import json
 
 from salt import exceptions
 import salt.utils.path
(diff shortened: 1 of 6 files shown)


salt/states/crmshmod.py
```diff
@@ -395,3 +405,146 @@ def corosync_updated(
     ret['comment'] = 'Corosync configuration file updated'
     ret['result'] = True
     return ret
+
+
+def cluter_properties_present(
```
Simranpal, line 410: `cluter/ cluster` in function name

salt/states/crmshmod.py
```diff
@@ -395,3 +405,146 @@ def corosync_updated(
...
+        __salt__['crm.configure_rsc_defaults'](option=option, value=value)
+        changes[option] = value
+
+    ret['changes'] = changes
+    ret['comment'] = 'Cluster rsc_defaults configured'
+    ret['result'] = True
+    return ret
+
+
+def cluter_op_defaults_present(
```
Simranpal, line 508: `cluter/ cluster` in function name

salt/states/crmshmod.py
```diff
@@ -395,3 +405,146 @@ def corosync_updated(
...
+        ret['result'] = False
+        return ret
+
+    ret['changes'] = changes
+    ret['comment'] = 'Cluster properties configured'
+    ret['result'] = True
+    return ret
+
+
+def cluter_rsc_defaults_present(
```
Simranpal, line 463: `cluter/ cluster` in function name

salt/modules/crmshmod.py
```diff
@@ -577,20 +579,95 @@ def configure_load(
         Used method (check in the description)
     url
         Used configuration file url (or path if it's a local file)
-    is_xml:
+    is_xml
```
Simranpal, line 582: I noticed the semicolon` : ` is used irregularly here for docstring values for all functions, or is there a pattern to follow

salt/modules/crmshmod.py
```diff
@@ -577,20 +579,95 @@ def configure_load(
...
     CLI Example:
 
     .. code-block:: bash
 
         salt '*' crm.configure_load update file.conf
     '''
-    cmd = '{crm_command} configure load {xml}{method} {url}'.format(
+    cmd = '{crm_command} {force} configure load {xml}{method} {url}'.format(
         crm_command=CRM_COMMAND,
+        force='-F' if force else '-n',
```
Simranpal, line 595: what is the `-n` option for

salt/modules/crmshmod.py
arbulu89: It shouldn't be set. It's not used in salt documentation. Anyway, we will find it here and there...
Not a big deal, if we find, just try to remove it.

salt/modules/crmshmod.py
arbulu89:
```
-F, --force           Make crm proceed with applying changes where it would
                        normally ask the user to confirm before proceeding.
                        This option is mainly useful in scripts, and should be
                        used with care.
  -n, --no              Automatically answer no when prompted```

arbulu89:
@Simranpal Changes done. Have a look again when you can. Thanks!

PR #65, opened 2020-06-05
Retry get json status when estimation sync time not calculated yet.

nick-wang:
Estimation sync time still need to be calculated based
on the synced data. Though less than a sec,
`drbdsetup status --json` will return an invalid json
format `"estimated-seconds-to-finish": nan,` when not ready.

### DIFF
--- a/salt/modules/drbdmod.py
+++ b/salt/modules/drbdmod.py
@@ -17,6 +17,7 @@ Module to provide DRBD functionality to Salt
 from __future__ import absolute_import, print_function, unicode_literals
 
 import logging
+import time
 
 from salt.exceptions import CommandExecutionError
 from salt.ext import six
@@ -36,6 +37,9 @@ WITH_JSON = True
 DRBDADM = 'drbd-utils'
 # drbd-utils >= 9.0.0 for json status
 DRBDADM_JSON_VERSION = '9.0.0'
+
+RETRY_TIMES = 5
+RETRY_INTERVAL = 3
 
 
 def __virtual__():  # pragma: no cover
@@ -278,6 +282,35 @@ def _is_no_backing_dev_request(res, output):
             return False
 
     return True
+
+
+def _get_json_output_save(command):
+    '''
+    '''
+    attempts = 0
+
+    while attempts < RETRY_TIMES:
+
+        try:
+            results = __salt__['cmd.run_all'](command)
+
+            if 'retcode' not in results or results['retcode'] != 0:
+                LOGGER.info('No cmd(%s) due to %s (%s).', command,
+                            results['stderr'], results['retcode'])
+                return None
+
+            ret = salt.utils.json.loads(results['stdout'], strict=False)
+            break
+
+        except ValueError:
+            attempts += 1
+            time.sleep(RETRY_INTERVAL)
+
+    else:
+        raise CommandExecutionError('Error happens when try to load the json output.',
+                                    info=results)
+
+    return ret
 
 
 def overview():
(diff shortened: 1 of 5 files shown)


nick-wang:
Besides this change in drbd salt module, also working in progress on fix in drbd-utils. Though the later one will take longer time to upstream and back.

nick-wang:
The fix in upstream https://github.com/LINBIT/drbd-utils/commit/104293030b2c0106b4791edb3eec38b476652a2e is the reason why this issue not shown in SLE15.

This commit doesn't need to be revert even the patch backported to SLE12SP4/SP5. Also this commit also helpful during the time of patch release in maintenance channel, which maybe quite long...

salt/modules/drbdmod.py
```diff
@@ -280,6 +284,35 @@ def _is_no_backing_dev_request(res, output):
     return True
 
 
+def _get_json_output_save(command):
+    '''
```
arbulu89, line 288: Add the reason why we are creating this issue here in the docstring. I added some context in the main review comment

arbulu89:
> The fix in upstream [LINBIT/drbd-utils@1042930](https://github.com/LINBIT/drbd-utils/commit/104293030b2c0106b4791edb3eec38b476652a2e) is the reason why this issue not shown in SLE15.
> 
> This commit doesn't need to be revert even the patch backported to SLE12SP4/SP5. Also this commit also helpful during the time of patch release in maintenance channel, which maybe quite long...

@nick-wang Is there any plan to backport this to SLE12? This would avoid having the need to have workarounds in our salt code

nick-wang:
> @nick-wang Is there any plan to backport this to SLE12? This would avoid having the need to have workarounds in our salt code

@arbulu89 WIll backport the upstream fix to SLE12SP4/SP5, but still need this workaround before patch released.
Change the loop to context convert in the latest commit.

salt/modules/drbdmod.py
```diff
@@ -280,6 +280,37 @@ def _is_no_backing_dev_request(res, output):
...
+    A warpper of get json command to acommandate json output issues
+    '''
+
+    error_str = '"estimated-seconds-to-finish": nan,'
+    replace_str = '"estimated-seconds-to-finish": 987654321,'
```
arbulu89, line 289: @nick-wang Why don't you convert the value to `null` ? In fact, `nan` is used for that. I think It makes sense to have as close as possible output as the final solution.

salt/modules/drbdmod.py
```diff
@@ -280,6 +280,37 @@ def _is_no_backing_dev_request(res, output):
...
+
+    if 'retcode' not in results or results['retcode'] != 0:
+        LOGGER.info("Error running command \"%s\".  Error message: %s (%s)",
+                     command, results['stderr'], results['retcode'])
+        return None
+
+    # drbd-utils 104293030b2c0106b4791edb3eec38b476652a2e
+    # results['stdout'] is unicode
+    s_result = str(results['stdout'])
+    if error_str in s_result:
```
arbulu89, line 300: I think you can skip this `if` here. The `replace` will only take place if the 1st pattern is present. Anyway, It's fine as it is too.

salt/modules/drbdmod.py
nick-wang: @arbulu89 I use `987654321` just match to the upstream change in https://github.com/LINBIT/drbd-utils/commit/104293030b2c0106b4791edb3eec38b476652a2e
, in which is use the number. 
Considering we use the function to check sync finished or not, the number is actually meaningless.

salt/modules/drbdmod.py
nick-wang: Only use it to save/skip the step of `results['stdout'] = six.text_type(s_result)`.
It won't be performed in most cases, without a replacement, the results['stdout'] will keep the same.

PR #66, opened 2020-06-09
hana: Fix imp reload usage in py2

arbulu89:
The `reload_hdb_connector` method in py2 was not working, as the `importlib.reload` method doesn't exist.
Now the code imports the correct reload from `importlib/imp` depending on the python version.
I will run tests now for py2 and py3.

### DIFF
--- a/_service
+++ b/_service
@@ -4,7 +4,7 @@
     <param name="scm">git</param>
     <param name="exclude">.git</param>
     <param name="filename">salt-shaptools</param>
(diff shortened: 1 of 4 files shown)


arbulu89:
You were too fast merging @stefanotorresi hehe
I have not finished testing everything yet, and maybe there is some issue. My fault anyway, I should've set in draft first

stefanotorresi:
gah, sorry! yep, I guess draft is made exactly for that purpose.

PR #70, opened 2020-09-11
Create cloud_grains_present state

arbulu89:
New salt state for https://github.com/SUSE/habootstrap-formula/pull/58

It basically assigns the required cloud information to grains to be used later. This is the unique way to use runtime variables in salt in the same execution.

Fix for: https://github.com/SUSE/habootstrap-formula/issues/57

### DIFF
--- a/salt/states/crmshmod.py
+++ b/salt/states/crmshmod.py
@@ -548,3 +548,50 @@ def cluster_op_defaults_present(
     ret['comment'] = 'Cluster op_defaults configured'
     ret['result'] = True
     return ret
+
+
+def cloud_grains_present(
+        name):
+    """
+    Set the required cloud providers data in the grains
+
+    All providers: cloud_provider
+    Only gcp: gcp_instance_id, gcp_instance_name
+
+    name:
+        This parameter is ignored
+    """
+
+    changes = {}
+    ret = {'name': name,
+           'changes': changes,
+           'result': False,
+           'comment': ''}
+
+    if __opts__['test']:
+        ret['result'] = None
+        ret['comment'] = 'Cloud grains would be set'
+        ret['changes'] = changes
+        return ret
+
+    cloud_provider = __salt__['crm.detect_cloud']()
+    __salt__['grains.set']('cloud_provider', cloud_provider)
+    changes['cloud_provider'] = cloud_provider
+
+    if cloud_provider == 'google-cloud-platform':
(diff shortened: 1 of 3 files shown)


MalloZup (commented):
to me it looks more complicating the code instead of solving issue.

I'm referring on the comment I left on https://github.com/SUSE/habootstrap-formula/issues/57

arbulu89:
@MalloZup It's not more complicated. We had the same `http` calls before, exactly the same. The unique difference is that the code execution has been moved to salt states instead of `jinja` code, what is not really recommended.
Salt code in `sls` formats has some limitations, and due them, we are forced to create code in `python`. I would like to have this implementation in the formula too, but it's not possible as far I have researched.

MalloZup:
@arbulu89  yes I checked  the http request is moved. make sense WIll check during the day

salt/states/crmshmod.py
```diff
@@ -548,3 +548,50 @@ def cluster_op_defaults_present(
     ret['comment'] = 'Cluster op_defaults configured'
     ret['result'] = True
     return ret
+
+
+def cloud_grains_present(
+        name):
+    """
+    Set the required cloud providers data in the grains
```
MalloZup, line 556: can we add why this is needed on highlevel manner?
why do we need to set them?

MalloZup (approved):
ok only minor comment on UX

salt/states/crmshmod.py
```diff
@@ -548,3 +548,50 @@ def cluster_op_defaults_present(
     ret['comment'] = 'Cluster op_defaults configured'
     ret['result'] = True
     return ret
+
+
+def cloud_grains_present(
```
Simranpal, line 553: should this not be named like `set_cloud_grains`

Simranpal (approved):
Thanks @arbulu89 just a comment about naming. This approach is aligns better than jinja salt calls

salt/states/crmshmod.py
arbulu89: Why do you prefer `set_cloud_grains` over `cloud_grains_present`?
This is a salt state, and as convention, the states are named with this syntax, saying which will be the state of the provisioning after the execution.
If you check the rest of the file, the rest of the methods are following the same syntax rules.

Simranpal: yes, you are right, I forgot this was salt state name

PR #72, opened 2020-09-24
Feature/ensa grains present

arbulu89:
Implement a new state to store the currently installed ENSA version. Thje new state sets the `ensa_version` grain with after being executed.
This state will be used in `sapnwbootstrap-formula`

Based on:
https://github.com/SUSE/shaptools/pull/59

More information in:
https://www.suse.com/c/suse-ha-sap-certified/
https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse#2d6008b0-685d-426c-b59e-6cd281fd45d7
https://launchpad.support.sap.com/#/notes/1410736

FYI @petersatsuse

### DIFF
--- a/salt/states/netweavermod.py
+++ b/salt/states/netweavermod.py
@@ -490,8 +490,12 @@ def ensa_version_grains_present(
         ret['changes'] = changes
         return ret
 
-    ensa_version = __salt__['netweaver.get_ensa_version'](
-        sap_instance, sid, inst, password)
+    try:
+        ensa_version = __salt__['netweaver.get_ensa_version'](
+            sap_instance, sid, inst, password)
+    except exceptions.CommandExecutionError as err:
+        ret['comment'] = six.text_type(err)
+        return ret
 
     __salt__['grains.set']('ensa_version', ensa_version)
     changes['ensa_version'] = ensa_version
--- a/salt/states/netweavermod.py
+++ b/salt/states/netweavermod.py
@@ -464,7 +464,7 @@ def ensa_version_grains_present(
         inst=None,
         password=None):
     '''
-    Set the `ensa_version` grain with the currently installed ENSA version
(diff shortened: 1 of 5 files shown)


salt/states/netweavermod.py
```diff
@@ -456,3 +456,53 @@ def sapservices_updated(
...
+
+    try:
+        ensa_version = __salt__['netweaver.get_ensa_version'](
+            sap_instance, sid, inst, password)
+    except exceptions.CommandExecutionError as err:
+        ret['comment'] = six.text_type(err)
+        return ret
+
+    grain_key = 'ensa_version_{}_{}'.format(sid, inst)
+    __salt__['grains.set'](grain_key, ensa_version)
```
Simranpal, line 502: After this statement, does the grains or pillar needs refresh. I was looking at grains.setval which has that option to refresh pillar

https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.grains.html

Simranpal (approved):
Looks good! Just one comment about the grains set fn above

salt/states/netweavermod.py
MalloZup: I do think we might need also to use the `force` option since if the grains are already set , a 2nd run will not overwrite them.

So imho that can be source of errors.

MalloZup (commented):
I would think we need force.

Regarding `refresh` grains I dunno, it is something we need imho to search and maybe document it on some of our confluence page like best practices to deal with grains

salt/states/netweavermod.py
arbulu89: 1. No, we don't need any refresh. It works as it is.
2. We don't need `force`. This is a single value, not a list or dict. For this case force is not required and the value is always overwritten

PR #73, opened 2020-10-15
Hotfix/improve extract pydbapi

arbulu89:
Implement the improvement of `extract_dbapi` to check recursively in subfolders. This is needed in order to find the `PYDBAPI.TGZ` file in a SAP HANA client folder, if this is extracted from a `SAR` file.

This folder is extracted putting the HANA client in `IMDB_CLIENT20_004_182-80002082/SAP_HANA_CLIENT` kind of folder, so if we check directly in the 1st one, the file is not there, so we need to check in the child subfolder.

Besides that, many imported packages for Unit tests have been removed, as they are not shipped anymore in Salt 3000 version (`NO_MOCK` and `NO_MOCK_REASON`)

### DIFF
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -27,6 +27,7 @@ import logging
 import time
 import re
 import sys
+import os
 
 if sys.version_info.major == 2: # pragma: no cover
     import imp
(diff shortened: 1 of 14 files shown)


salt/modules/hanamod.py
```diff
@@ -959,9 +960,17 @@ def __import_mod(name):
...
+def _find_sap_folder(software_folders, folder_pattern, check_subfolder_level=0):
     '''
     Find a SAP folder following a recursive approach using the LABEL and LABELIDX files
+
+    Args:
+        software_folder (list): List of subfolder where the SAP folder is looked for`
+        folder_pattern (str): Pattern of the LABEL.ASC to look fo
+        check_subfolder_level (int): Number of subfolder levers to check.
+            Examples:
+                1 means to check recursively in the subfolder present in sofware_folder folders
```
MalloZup, line 972:
```suggestion
                1 means to check recursively in the subfolder present in software_folder folders
```

salt/modules/hanamod.py
```diff
@@ -959,9 +960,17 @@ def __import_mod(name):
...
+                2 measn to check in the first subfolder and the folder within
```
MalloZup, line 973:
```suggestion
                2 mean to check in the first subfolder and the folder within
```

salt/modules/hanamod.py
```diff
@@ -959,9 +960,17 @@ def __import_mod(name):
     reload_module(hdb_connector)
 
 
-def _find_sap_folder(software_folders, folder_pattern):
+def _find_sap_folder(software_folders, folder_pattern, check_subfolder_level=0):
```
MalloZup, line 963: I would rename `check_subfolder_level` to `recursion_level`

MalloZup (commented):
overall looks good but I would call the argument more recursion level seems more clear imho

MalloZup (approved):
thx

Simranpal (approved):
Thanks. I see we may use the same functionality for checking sub-folders for other sap components in future

PR #74, opened 2020-11-11
Add qdevice

aleksei-burlakov:
@arbulu89 @MalloZup @Simranpal  Please have a look. This PR should come together with https://github.com/SUSE/habootstrap-formula/pull/67

### DIFF
--- a/salt-shaptools.changes
+++ b/salt-shaptools.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Wed Dec 16 09:04:29 UTC 2020 - Aleksei Burlakov <aburlakov@suse.com>
+
+- qdevice support: it can be created when initializing a cluster
+
+-------------------------------------------------------------------
 Thu Oct 15 07:14:20 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
 
 - Version 0.3.11
--- a/salt/modules/crmshmod.py
+++ b/salt/modules/crmshmod.py
@@ -290,6 +290,7 @@ def _crm_init(
         sbd=None,
         sbd_dev=None,
         no_overwrite_sshkey=False,
+        qnetd_hostname=None,
         quiet=None):
     '''
     crm cluster init command execution
@@ -311,6 +312,9 @@ def _crm_init(
         cmd = '{cmd} -S'.format(cmd=cmd)
     if no_overwrite_sshkey:
         cmd = '{cmd} --no-overwrite-sshkey'.format(cmd=cmd)
+    if qnetd_hostname:
+        cmd = '{cmd} --qnetd-hostname {qnetd_hostname}'.format(
+            cmd=cmd, qnetd_hostname=qnetd_hostname)
     if quiet:
         cmd = '{cmd} -q'.format(cmd=cmd)
 
@@ -324,6 +328,7 @@ def _ha_cluster_init(
         admin_ip=None,
         sbd=None,
         sbd_dev=None,
+        qnetd_hostname=None,
         quiet=None):
     '''
     ha-cluster-init command execution
@@ -342,6 +347,9 @@ def _ha_cluster_init(
         cmd = '{cmd} {sbd_str}'.format(cmd=cmd, sbd_str=sbd_str)
     elif sbd:
         cmd = '{cmd} -S'.format(cmd=cmd)
+    if qnetd_hostname:
+        cmd = '{cmd} --qnetd-hostname {qnetd_hostname}'.format(
+            cmd=cmd, qnetd_hostname=qnetd_hostname)
     if quiet:
         cmd = '{cmd} -q'.format(cmd=cmd)
 
@@ -362,6 +370,7 @@ def cluster_init(
         sbd=None,
         sbd_dev=None,
         no_overwrite_sshkey=False,
+        qnetd_hostname=None,
         quiet=None):
     '''
     Initialize a cluster from scratch.
@@ -387,6 +396,8 @@ def cluster_init(
     no_overwrite_sshkey
         No overwrite the currently existing sshkey (/root/.ssh/id_rsa)
         Only available after crmsh 3.0.0
+    qnetd_hostname:
+        The name of the qnetd node. If none, no qdevice is created
     quiet:
         execute the command in quiet mode (no output)
 
@@ -403,13 +414,14 @@ def cluster_init(
     # and create the corresponing UT
     if __salt__['crm.use_crm']:
         return _crm_init(
-            name, watchdog, interface, unicast, admin_ip, sbd, sbd_dev, no_overwrite_sshkey, quiet)
+            name, watchdog, interface, unicast, admin_ip, sbd, sbd_dev, no_overwrite_sshkey,
+            qnetd_hostname, quiet)
 
     LOGGER.warning('The parameter name is not considered!')
     LOGGER.warning('--no_overwrite_sshkey option not available')
 
     return _ha_cluster_init(
-        watchdog, interface, unicast, admin_ip, sbd, sbd_dev, quiet)
+        watchdog, interface, unicast, admin_ip, sbd, sbd_dev, qnetd_hostname, quiet)
 
 
 def _crm_join(
--- a/salt/states/crmshmod.py
+++ b/salt/states/crmshmod.py
@@ -106,7 +106,8 @@ def cluster_initialized(
         sbd=None,
         sbd_dev=None,
         no_overwrite_sshkey=False,
-        quiet=None):
+        quiet=None,
+        qnetd_hostname=None):
     """
     Machine is running a cluster node
 
@@ -131,6 +132,8 @@ def cluster_initialized(
         Only available after crmsh 3.0.0
     quiet:
         execute the command in quiet mode (no output)
+    qnetd_hostname:
+        The name of the qnetd node. If none, no qdevice is created
     """
     ret = {'name': name,
            'changes': {},
@@ -159,6 +162,7 @@ def cluster_initialized(
             sbd=sbd,
             sbd_dev=sbd_dev,
             no_overwrite_sshkey=no_overwrite_sshkey,
+            qnetd_hostname=qnetd_hostname,
             quiet=quiet)
 
         if result:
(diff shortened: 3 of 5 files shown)


MalloZup (changes requested):
thx for pr ! 3 things:

1) CODE: the code itself need adaption with the existing code. see inline comments

2) TEST: you need to adapt the tests mocking. see travis failures. ( you can find documentation on mocking on the readme, and looking at existing tests)

3) PKG: Releases: You need to add  a changelog entry to
https://github.com/SUSE/salt-shaptools/blob/master/salt-shaptools.changes#L2

and also increment the version number so basically a new pkg will shipped and then the PR you depend on this will also get the new pkgs of this shared lib.

here is where you need to increment: 
https://github.com/SUSE/salt-shaptools/blob/master/_service#L7

salt/modules/crmshmod.py
```diff
@@ -352,6 +352,13 @@ def _ha_cluster_init(
         _set_corosync_unicast(addr, name)
     return return_code
 
+def _crm_init_qdevice(qnetd_hostname=None):
+    '''
+    crm cluster init qdevice command execution
+    '''
+    cmd = 'crm cluter init qdevice --qnetd-hostname={qnetd_hostname} -y'.format(
```
MalloZup, line 359: here use `{crm_command}`

salt/modules/crmshmod.py
```diff
@@ -352,6 +352,13 @@ def _ha_cluster_init(
         _set_corosync_unicast(addr, name)
     return return_code
 
+def _crm_init_qdevice(qnetd_hostname=None):
```
MalloZup, line 355: I think you need also to implement the `quite` parameter as the other functions

https://github.com/SUSE/salt-shaptools/pull/74/files#diff-4028dde1b1b110cd28ee9b7904381ce5b8f2666971150218ed60c08c0f6b7e61R431

salt/modules/crmshmod.py
MalloZup: https://github.com/SUSE/salt-shaptools/pull/74/files#diff-4028dde1b1b110cd28ee9b7904381ce5b8f2666971150218ed60c08c0f6b7e61R435

salt/modules/crmshmod.py
Simranpal, line 355: The function parameters are missing in the docstring for this function

salt/states/crmshmod.py
```diff
@@ -131,6 +132,8 @@ def cluster_initialized(
         Only available after crmsh 3.0.0
     quiet:
         execute the command in quiet mode (no output)
+    qnetd_hostname:
+        create the qdevice if defined
```
Simranpal, line 136: I would add some more description about the hostname, and thats its needed to setup the QDevice

salt/modules/crmshmod.py
Simranpal: typo `cluter/cluster`

Simranpal (changes requested):
@aleksei-burlakov some comments below

arbulu89 (commented):
@aleksei-burlakov As a side note, I think that we can initialize qdevice in the same command where we start the cluster. This could spare this new command, and we would only need to add the flags in the current code. Just as an additional option.

Please, add the args description in the docstrings too

aleksei-burlakov:
> @aleksei-burlakov As a side note, I think that we can initialize qdevice in the same command where we start the cluster. This could spare this new command, and we would only need to add the flags in the current code. Just as an additional option.
> 
> Please, add the args description in the docstrings too

Sure. First I wanted to do it like this but then found so many functions, that I decided that this is the style to put it all in a separate function.

salt/modules/crmshmod.py
```diff
@@ -399,18 +411,23 @@ def cluster_init(
     if sbd_dev and not isinstance(sbd_dev, list):
         sbd_dev = [sbd_dev]
 
+    LOGGER.warning('The parameter name is not considered!')
```
arbulu89, line 414: Does this have any relationship with `qdevice`?
It doesn't look like.
I would split this into other PR. Or at least other commit, as they are unrelated

salt/modules/crmshmod.py
```diff
@@ -352,6 +352,17 @@ def _ha_cluster_init(
         _set_corosync_unicast(addr, name)
     return return_code
 
+def _crm_init_qdevice(
+        quiet=None,
```
arbulu89, line 356: I would put `quiet` as last parameter. It looks secondary.
Usually, any user that wants to use this method only would put the `qnetid_hostname` parameter only, so being the first makes sense

salt/modules/crmshmod.py
```diff
@@ -399,18 +411,23 @@ def cluster_init(
...
+    else:
+        return_code = _ha_cluster_init(
+            watchdog, interface, unicast, admin_ip, sbd, sbd_dev, quiet)
 
-    LOGGER.warning('The parameter name is not considered!')
-    LOGGER.warning('--no_overwrite_sshkey option not available')
-
-    return _ha_cluster_init(
-        watchdog, interface, unicast, admin_ip, sbd, sbd_dev, quiet)
+    if qnetd_hostname:
```
arbulu89, line 427: Does it make any sense to run this 2nd operation if the `crm_init` failed?
Maybe, we just need to finish the process if that 1st one fails.
Something like `if not return_code and qnetd_hostname:`

arbulu89 (changes requested):
@aleksei-burlakov Added some comments.

After dealing with them, we will need to add/update the UT

salt/modules/crmshmod.py
aleksei-burlakov: Done.

aleksei-burlakov: Done.

salt/states/crmshmod.py
aleksei-burlakov: Done.

aleksei-burlakov:
> @aleksei-burlakov As a side note, I think that we can initialize qdevice in the same command where we start the cluster. This could spare this new command, and we would only need to add the flags in the current code. Just as an additional option.
> 
> Please, add the args description in the docstrings too

Done.

aleksei-burlakov:
> thx for pr ! 3 things:
> 
> 1. CODE: the code itself need adaption with the existing code. see inline comments
> 2. TEST: you need to adapt the tests mocking. see travis failures. ( you can find documentation on mocking on the readme, and looking at existing tests)
> 3. PKG: Releases: You need to add  a changelog entry to
>    https://github.com/SUSE/salt-shaptools/blob/master/salt-shaptools.changes#L2
> 
> and also increment the version number so basically a new pkg will shipped and then the PR you depend on this will also get the new pkgs of this shared lib.
> 
> here is where you need to increment:
> https://github.com/SUSE/salt-shaptools/blob/master/_service#L7

I have refactored the 1. CODE. No, only left are the 2. TEST and 3. Changelog

salt/modules/crmshmod.py
aleksei-burlakov: Done.

salt/modules/crmshmod.py
aleksei-burlakov: Done.

salt/modules/crmshmod.py
aleksei-burlakov: It's a single command now.

arbulu89: @aleksei-burlakov Remember that this has the inconvenience that if the cluster is already running, it won't add the `qdevice` part there. We need to have this tracked down somewhere, as it's important to keep the state as most idempotent as possible.
Scenarios:
- Cluster is not initialized. This will create the cluster and add qdevice
- Cluster is initialized with qdevice. Most probably nothing is changed
- Cluster is initialized **without qdevice**. Nothing is changed, and this is wrong, as I would expect that `qdevice` is added

aleksei-burlakov:
I have also added the tests.

salt/modules/crmshmod.py
aleksei-burlakov: > @aleksei-burlakov Remember that this has the inconvenience that if the cluster is already running, it won't add the `qdevice` part there. We need to have this tracked down somewhere, as it's important to keep the state as most idempotent as possible.
> Scenarios:
> 
> * Cluster is not initialized. This will create the cluster and add qdevice
> * Cluster is initialized with qdevice. Most probably nothing is changed
> * Cluster is initialized **without qdevice**. Nothing is changed, and this is wrong, as I would expect that `qdevice` is added

This one is outdated. The cluster is initialized in one-shot now. $ crm cluster init -s /dev/sbd ... --qnetd-hostname -y

aleksei-burlakov: It makes sense.

aleksei-burlakov: I see the sbd device and other parameters are also not altered on an existing running cluster when we try to initialize a new one. If we want to do this, it will require changing the old code, so I propose to split it into another PR. What do you think @arbulu89, @MalloZup, @stefanotorresi ?
I will create an issue in the meanwhile.

aleksei-burlakov: I've created one there https://github.com/SUSE/salt-shaptools/issues/76

arbulu89: Ok, fair point. If we track this somewhere else we can split in a future implementation.
Could you paste the comments I did in that new ticket?

arbulu89 (approved):
@aleksei-burlakov The changes and the tests look good.
Add a changelog entry with `osc vc` explaining `qdevice` is supported now, and we are ready to go.
I'm approving the PR to avoid more reviews, but take care of the changes file first please.

aleksei-burlakov:
@MalloZup , @Simranpal please review the last changes.

salt-shaptools.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Dec 16 09:04:29 UTC 2020 - Aleksei Burlakov <aburlakov@suse.com>
+
+- Version 0.3.12
```
arbulu89, line 4: @aleksei-burlakov Please, remove the new version information.
Add just the next text.
We will add the version when we want to release a new package, most probably with more changes. (Like the one coming in https://github.com/SUSE/salt-shaptools/pull/77 and others).

Simranpal (approved):
thanks for the changes @aleksei-burlakov It looks good now

salt-shaptools.changes
aleksei-burlakov: @arbulu89 I just followed the https://github.com/SUSE/salt-shaptools/pull/74#pullrequestreview-528936398. I will remove the version change.

aleksei-burlakov: @arbulu89 the version bump is reverted.

PR #77, opened 2020-12-16
Remove --no-overwrite-sshkey: the option is deprecated

aleksei-burlakov:
The --no-overwrite-sshkey is deprecated. There are 2 commits in this PR, but only the last is relevant. The first commit will disappear when #74 is merged.

Jira reference: https://jira.suse.com/browse/TEAM-894

### DIFF
--- a/salt-shaptools.changes
+++ b/salt-shaptools.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Wed Dec 16 09:04:29 UTC 2020 - Aleksei Burlakov <aburlakov@suse.com>
+
+- qdevice support: it can be created when initializing a cluster
+
+-------------------------------------------------------------------
 Thu Oct 15 07:14:20 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
 
 - Version 0.3.11
--- a/salt-shaptools.changes
+++ b/salt-shaptools.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Wed Dec 16 11:30:22 UTC 2020 - Aleksei Burlakov <aburlakov@suse.com>
+
+- remove --no-overwrite-sshkey: the option is deprecated
+
+-------------------------------------------------------------------
 Wed Dec 16 09:04:29 UTC 2020 - Aleksei Burlakov <aburlakov@suse.com>
 
 - qdevice support: it can be created when initializing a cluster
--- a/tests/unit/states/test_crmshmod.py
+++ b/tests/unit/states/test_crmshmod.py
@@ -189,6 +189,7 @@ class CrmshmodTestCase(TestCase, LoaderModuleMockMixin):
                 sbd=True,
                 sbd_dev='/dev/sbd',
                 no_overwrite_sshkey=True,
+                qnetd_hostname=None,
                 quiet=False)
 
     def test_initialized_error(self):
@@ -225,6 +226,7 @@ class CrmshmodTestCase(TestCase, LoaderModuleMockMixin):
                 sbd=True,
                 sbd_dev='/dev/sbd',
                 no_overwrite_sshkey=True,
+                qnetd_hostname=None,
                 quiet=False)
 
     def test_initialized_command_error(self):
@@ -262,6 +264,7 @@ class CrmshmodTestCase(TestCase, LoaderModuleMockMixin):
                 sbd=True,
                 sbd_dev='/dev/sbd',
                 no_overwrite_sshkey=False,
+                qnetd_hostname=None,
                 quiet=False)
 
     # 'joined' function tests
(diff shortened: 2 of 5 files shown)


salt-shaptools.changes
```diff
@@ -1,3 +1,15 @@
+-------------------------------------------------------------------
+Wed Dec 16 11:30:22 UTC 2020 - Aleksei Burlakov <aburlakov@suse.com>
+
+- Version 0.3.13
```
arbulu89, line 4: Don't increase the version number. Just add the changed functionality test.
We will increase the version once we want to publish the new changes

arbulu89 (commented):
@aleksei-burlakov To remove the command as you did, we would need to publish a new major version, as it is not backward compatible. I see these options:
1. We release a new `0.4.0` version with this change (when we want to publish to SLE)
2. Instead of removing the option, we just send a warning message saying this variable is deprecated

If we take option 1 *which might make sense), we can continue with these exact changes (except the version info in the changes file)

salt-shaptools.changes
aleksei-burlakov: We were already warning before. IMHO the first option is better from all sides. If we choose the first option, I would add another change, namely, I would change the order of the arguments in here https://github.com/SUSE/salt-shaptools/pull/74/commits/a3e16de2380f3255e92c5e6e1b5c4f28acc52e8b#diff-924d6f55de38bd356ee8e2823ed4fad0cbaa34a1cccf7dff0d8b98b0cc3dec30R109-R110
The argument `qnetd_hostname` comes before `quiet` in the `modules/crmshmod.py` in #74. It's what we want. However the `states/crmshmod.py:cluster_initialized` is used outside, so I put the `qnetd_hostname` as the last parameter. To make it consistent, the `quiet` parameter should be the last.
So, I would replace `quiet` and `qnetd_hostname` in `cluster_initialized` in this PR and set the version to 0.4.0
@arbulu89 what do you think?

tests/unit/states/test_crmshmod.py
```diff
@@ -188,7 +187,7 @@ def test_initialized(self):
                 admin_ip='192.168.1.50',
                 sbd=True,
                 sbd_dev='/dev/sbd',
-                no_overwrite_sshkey=True,
+                qnetd_hostname=None,
```
Simranpal, line 190: The order for arguments `quiet` and `qnetd_hostname` should be same as in the original function

Simranpal (approved):
@aleksei-burlakov looks good. About the order of arguments, I think it will be better if its consistent between the related module and state functions

arbulu89:
@aleksei-burlakov As commented before. Please, don't increase the version number in the changelog. We will do that once we want to release, as we will get many changes at once.

salt-shaptools.changes
MalloZup: is this comment on the righ place?

MalloZup (changes requested):
I agree about the `version` number with xabi.

I would also like to have in the `PR` Description the motivation/rationale why the option `no-overwrite` is becoming depracted.

A part of that , looks good to me . Good work!

salt-shaptools.changes
aleksei-burlakov: @MalloZup , Yes it's on it's place, but I see it's not really obvious what I say there. Xiabi gives two options. First is the version increase, the second is to warn the user. I give arguments about why the first option is better and there come the details. To cut a long story short, because it won't be backward compatible. But, I understand we will increase the version number on the next release.

tests/unit/states/test_crmshmod.py
aleksei-burlakov: @Simranpal there you see the change from two commits. First introduces the qnetd_hostname and the second removes the overwrite_sshkey. I followed the advice from @arbulu89 https://github.com/SUSE/salt-shaptools/pull/74#discussion_r539087231 to make the quiet parameter the last, and to my mind, it's fair to make the quiet parameter the last.

aleksei-burlakov:
> I agree about the `version` number with xabi.
> 
> I would also like to have in the `PR` Description the motivation/rationale why the option `no-overwrite` is becoming depracted.
> 
> A part of that , looks good to me . Good work!

**The motivation** to remove the `no-overwrite` parameter is first because there is no need for it, second because it doesn't work. The first is somewhat speculative, one may say he really wants to have the cluster with the new ssh keys, though I can hardly imagine such a scenario. Initially, I guess, this argument was added just because this was the way how the `crm cluster init` worked. Now the `crm` doesn't overwrite the keys and I think this is very good.
Second, there is a functionality in the habootstrap-formula that is supposed to overwrite the ssh keys using this option, but it didn't work properly (see https://github.com/SUSE/habootstrap-formula/pull/68#issuecomment-750064240)

aleksei-burlakov:
@arbulu89 @MalloZup the version bump is reverted.

PR #79, opened 2021-03-11
Add OCFS2 device support to cluster init

cbosdo:
In order to help creating clusters with OCFS2 device, add an `ocfs2_dev` parameter to the `crmshmod` initialization state and module.


salt/modules/crmshmod.py
```diff
@@ -413,15 +422,17 @@ def cluster_init(
     # INFO: 2 different methods are created to make easy to read/understand
     # and create the corresponing UT
     if __salt__['crm.use_crm']:
-        return _crm_init(
+        return_code = _crm_init(
```
arbulu89, line 425: Is there any reason for this change? Add the end you are returning always after the command
It is not neither good or bad, just curiosity hehe

arbulu89 (approved):
Hi @cbosdo ,
The implementation looks good.
As far as I remember the ocfs2 configuration must always be configured in a 2nd command, right? (we talked about this but I already forgot it hehe)

The unique potential issue I see with this is the possibility to not run the 2 commands always in the same sequence if a 3rd node is trying to join the cluster. Let me explain:
- crmsh has an internal mechanism which creates a lock to run the whole init operation atomically. If other join wants to join, it will wait until the whole operation is finished, and after that it will acquire the lock. At this point, if the join operations is faster than the new `ocfs2` initialization, the join will be executed first, and afterwards the this new init. I don't know if this can affect somehow the cluster. I don't think so, but I just wanted to expose some side effects of having the initialization in 2 different actions.

Any opinion on this @xinliang?

liangxin1300:
To be honest, crm cluster init with ocfs2 device, this functionality, was lack of testing for quite a long time
I need to revisit this scenarios and related codes

liangxin1300:
Related PR on crmsh side https://github.com/ClusterLabs/crmsh/pull/798
I did refactor for configuring OCFS2

liangxin1300:
> Hi @cbosdo ,
> The implementation looks good.
> As far as I remember the ocfs2 configuration must always be configured in a 2nd command, right? (we talked about this but I already forgot it hehe)
> 
> The unique potential issue I see with this is the possibility to not run the 2 commands always in the same sequence if a 3rd node is trying to join the cluster. Let me explain:
> 
> * crmsh has an internal mechanism which creates a lock to run the whole init operation atomically. If other join wants to join, it will wait until the whole operation is finished, and after that it will acquire the lock. At this point, if the join operations is faster than the new `ocfs2` initialization, the join will be executed first, and afterwards the this new init. I don't know if this can affect somehow the cluster. I don't think so, but I just wanted to expose some side effects of having the initialization in 2 different actions.
> 
> Any opinion on this @xinliang?

@arbulu89 There is no any configuration file need to write and sync for ocfs2, just adding related resources and put them in a group and then put that group into a clone on init node.

BTW, I refactored ocfs2 part on crmsh side, PR is in https://github.com/ClusterLabs/crmsh/pull/798
And the package for testing is in https://build.opensuse.org/package/show/home:XinLiang:branches:crmsh_ocfs2/crmsh
With that change, you can still using `vgfs` stage, but it's an alias of `ocfs2` stage
@cbosdo You could try to use it and any feedback are welcome!

cbosdo:
> > Hi @cbosdo ,
> > The implementation looks good.
> > As far as I remember the ocfs2 configuration must always be configured in a 2nd command, right? (we talked about this but I already forgot it hehe)
> > The unique potential issue I see with this is the possibility to not run the 2 commands always in the same sequence if a 3rd node is trying to join the cluster. Let me explain:
> > 
> > * crmsh has an internal mechanism which creates a lock to run the whole init operation atomically. If other join wants to join, it will wait until the whole operation is finished, and after that it will acquire the lock. At this point, if the join operations is faster than the new `ocfs2` initialization, the join will be executed first, and afterwards the this new init. I don't know if this can affect somehow the cluster. I don't think so, but I just wanted to expose some side effects of having the initialization in 2 different actions.
> > 
> > Any opinion on this @xinliang?
> 
> @arbulu89 There is no any configuration file need to write and sync for ocfs2, just adding related resources and put them in a group and then put that group into a clone on init node.
> 
> BTW, I refactored ocfs2 part on crmsh side, PR is in [ClusterLabs/crmsh#798](https://github.com/ClusterLabs/crmsh/pull/798)
> And the package for testing is in https://build.opensuse.org/package/show/home:XinLiang:branches:crmsh_ocfs2/crmsh
> With that change, you can still using `vgfs` stage, but it's an alias of `ocfs2` stage
> @cbosdo You could try to use it and any feedback are welcome!

Just pushed a new version of the PR using your new `crmsh` code if available

salt/modules/crmshmod.py
cbosdo: surely a leftover from a previous attempt. Fixed it back to direct returns

stefanotorresi:
Thanks again, @cbosdo!

PR #80, opened 2021-03-14
Improve crosync configure

liangxin1300:
#### To address these issues
* Parse multi interface/node as list in the dict
* When there is no space between words in corosync.conf, like "totem{" or "ring0_addr:10.10.10.121",
  corosync still works, but our parse results are wrong


salt/states/crmshmod.py
```diff
@@ -298,11 +299,16 @@ def cluster_configured(
         return ret
 
 
-def _convert2dict(file_content_lines):
+DUMMY_SEC_NAME_TEMPL = "__{}_list"
+KNOWN_SEC_NAMES_WITH_LIST = ("totem.interface", "nodelist.node")
```
liangxin1300, line 303: Predefined the section names possible with multi items

salt/states/crmshmod.py
```diff
@@ -313,61 +319,97 @@ def _convert2dict(file_content_lines):
...
     for key, value in corodict.items():
-        if isinstance(value, dict):
-            output += '{}{} {{\n'.format(indentation, key)
-            indentation += '\t'
-            output += _convert2corosync(value, indentation)
-            indentation = indentation[:-1]
-            output += '{}}}\n'.format(indentation)
-        else:
-            output += '{}{}: {}\n'.format(indentation, key, value)
+        output += _unpack_dict(key, value, indentation)
```
liangxin1300, line 404: Refactor this part to pass the codeclimate checking

salt/states/crmshmod.py
```diff
@@ -313,61 +319,97 @@ def _convert2dict(file_content_lines):
...
-        current_path = '{}.{}'.format(initial_path, key)
+        current_path = '{}.{}'.format(initial_path, key) if initial_path else key
         if key in main_dict.keys() and not isinstance(value, dict):
             if str(main_dict[key]) != str(value):
                 applied_changes[current_path] = value
             main_dict[key] = value
         elif key in main_dict.keys():
-            modified_dict, new_changes = _mergedicts(main_dict[key], value, applied_changes, current_path)
-            main_dict[key] = modified_dict
-            applied_changes.update(new_changes)
```
liangxin1300, line 341: From observing results, seems no need to update main_dict and applied_changes here.

liangxin1300:
@arbulu89 Please review this, thank you!

salt/states/crmshmod.py
```diff
@@ -313,61 +319,97 @@ def _convert2dict(file_content_lines):
...
+                    corodict[dummpy_sec_name] = []
+                corodict[dummpy_sec_name].append({sec_name: sub_dict})
+            else:
+                corodict[sec_name] = sub_dict
             index = i + new_index
-        elif line_items[0][-1] == ':':
-            corodict[line_items[0][:-1]] = line_items[-1]
+            initial_path = re.sub("\.{}".format(sec_name), "", initial_path) if "." in initial_path else ""
+        elif ':' in stripped_line:
+            key, *values = stripped_line.split(':')
```
arbulu89, line 336: I don't really like this. I guess you do this packing to filter the IP like values. Is there some better way to find those options? Like some regular expression. It looks pretty strange otherwise (it took me some time to understand what was going on here)

salt/states/crmshmod.py
```diff
@@ -313,61 +319,97 @@ def _convert2dict(file_content_lines):
...
             continue
 
-        line_items = stripped_line.split()
         if '{' in stripped_line:
-            corodict[line_items[0]], new_index = _convert2dict(file_content_lines[i+1:])
+            sec_name = re.sub("\s*{", "", stripped_line)
+            initial_path += ".{}".format(sec_name) if initial_path else sec_name
+            sub_dict, new_index = _convert2dict(file_content_lines[i+1:], initial_path)
+            if initial_path in KNOWN_SEC_NAMES_WITH_LIST:
+                dummpy_sec_name = DUMMY_SEC_NAME_TEMPL.format(sec_name)
```
arbulu89, line 327: I don't really like the `DUMMY` term. This value is actually a modified value, right?
Maybe better `modified_sec_name` ?

salt/states/crmshmod.py
arbulu89: Nice touch!

arbulu89 (approved):
@liangxin1300 
The change looks good (even though it is quite complex hehe, but I understand that this complexity is needed).
I have some some superficial comments, so feel free to change or not.
Really nice job!

salt/states/crmshmod.py
liangxin1300: Maybe we could add a comment here to explain the purpose
Regular expression might more prone to errors

salt/states/crmshmod.py
liangxin1300: Changed

liangxin1300:
> @liangxin1300
> The change looks good (even though it is quite complex hehe, but I understand that this complexity is needed).
> I have some some superficial comments, so feel free to change or not.
> Really nice job!

@arbulu89 Thanks for review!
BTW, we could discuss how to reuse this part of codes between crmsh and this project.
I mean, current configuration parser in `crmsh/corosync.py` cannot support future corosync3 well, and also has the issue to parse multi node in nodelist, and to support corosync3, we need quite a lot of refactors on bootstrap, a well design and well tested new parser will be a good infrastructure for those job

arbulu89:
> > @liangxin1300
> > The change looks good (even though it is quite complex hehe, but I understand that this complexity is needed).
> > I have some some superficial comments, so feel free to change or not.
> > Really nice job!
> 
> @arbulu89 Thanks for review!
> BTW, we could discuss how to reuse this part of codes between crmsh and this project.
> I mean, current configuration parser in `crmsh/corosync.py` cannot support future corosync3 well, and also has the issue to parse multi node in nodelist, and to support corosync3, we need quite a lot of refactors on bootstrap, a well design and well tested new parser will be a good infrastructure for those job

@liangxin1300 And this code supports corosync3?
Making this code available on `crmsh` looks difficult. The best option as some of the next steps would be to translate this to `crmsh` and import it here. The issue is that we discarded to import `crmsh` modules here to avoid issues with py2/py3 versioning.
We might need to have a 2nd look. If the code goes in the exact same way to all the `crmsh` versions we should have any issue, but we should test with both py2 and py3.

We can do the next:
1. We merge this PR
2. We copy/paste/adapt this code in crmsh to do the same thing
3. We do new changes in this project. Import crmsh and use the same commands here
**Final thing**: Could you add a new changelog entry in this project (`osc vc`)? To have the changes tracked. You don't have to upgrade the version. Just add a basic text in a new entry in the changes file

liangxin1300:
> > > @liangxin1300
> > > The change looks good (even though it is quite complex hehe, but I understand that this complexity is needed).
> > > I have some some superficial comments, so feel free to change or not.
> > > Really nice job!
> > 
> > 
> > @arbulu89 Thanks for review!
> > BTW, we could discuss how to reuse this part of codes between crmsh and this project.
> > I mean, current configuration parser in `crmsh/corosync.py` cannot support future corosync3 well, and also has the issue to parse multi node in nodelist, and to support corosync3, we need quite a lot of refactors on bootstrap, a well design and well tested new parser will be a good infrastructure for those job
> 
> @liangxin1300 And this code supports corosync3?

Not yet.
But support corosync3 is on the way, crmsh will create new branch to maintain SLE15+ codes, while using master branch to support corosync3.
I want to refactor the corosync parser to support parse corosync3 well, then backport this big changes into SLE15+ branch

> Making this code available on `crmsh` looks difficult. The best option as some of the next steps would be to translate this to `crmsh` and import it here. The issue is that we discarded to import `crmsh` modules here to avoid issues with py2/py3 versioning.

But in the future, might be after 30 Jun 2022, when 12sp3 end of life, we ha stack could not consider py2 anymore, right?

> We might need to have a 2nd look. If the code goes in the exact same way to all the `crmsh` versions we should have any issue, but we should test with both py2 and py3.
> 
> We can do the next:
> 
> 1. We merge this PR
> 2. We copy/paste/adapt this code in crmsh to do the same thing
> 3. We do new changes in this project. Import crmsh and use the same commands here

Agree with you.
Thanks @arbulu89 !

>    **Final thing**: Could you add a new changelog entry in this project (`osc vc`)? To have the changes tracked. You don't have to upgrade the version. Just add a basic text in a new entry in the changes file

Like this 9ce19d900cd1ceaa9aca6366267552d83de644ea?

arbulu89:
> Like this [9ce19d9](https://github.com/SUSE/salt-shaptools/commit/9ce19d900cd1ceaa9aca6366267552d83de644ea)?

Yes! Thanks

arbulu89:
@liangxin1300 By the way, I don't know why the UT execution has been skipped... Most probably these changes will break the tests. Let me check this

arbulu89:
@liangxin1300 I have re-enabled the test in the CI in: https://github.com/SUSE/salt-shaptools/pull/81
Let's wait until this PR is merged first to see what happens with the UT

liangxin1300:
> @liangxin1300 I have re-enabled the test in the CI in: #81
> Let's wait until this PR is merged first to see what happens with the O

liangxin1300:
> @liangxin1300 I have re-enabled the test in the CI in: #81
> Let's wait until this PR is merged first to see what happens with the UT

ok, I will complete UT code if needed

liangxin1300:
@arbulu89 The UT codes have passed:)

arbulu89 (approved):
@liangxin1300 Great work! I think we are ready to merge

PR #82, opened 2021-07-21
fix typo to fix uninstalled state

yeoldegrove:


yeoldegrove:
@arbulu89 please review and merge

salt-shaptools.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Wed Jul 21 07:55:12 UTC 2021 - Eike Waldt <waldt@b1-systems.de>
+
+- fix typo to fix uninstalled state
```
arbulu89, line 4: If you are going to increase the version number, specify it in the changes here please

arbulu89 (approved):
Thanks @yeoldegrove ,
It's a pity that we don't have the correct asserts in the tests... 
This should be easy to catch

salt-shaptools.changes
yeoldegrove: @arbulu89 done

arbulu89:
@yeoldegrove , Could you check why the CI tests are failling?

PR #83, opened 2021-08-30
switch to ubuntu-20.04 for tests to fix salt build

yeoldegrove:
- switch to ubuntu-20.04 for tests to fix salt build
- use python modules based on python version
- use opensuse/salt branch based on python version


yeoldegrove:
@arbulu89 Please approve the workflow.

arbulu89 (approved):
Nice!

PR #84, opened 2022-01-21
add module to query remote HANA

yeoldegrove:
This adds a module to query a remote HANA DB.  This let's us interact with the HANA from e.g. a S/4HANA PAS server.

It adds countless new possibilities.
- execute backups on a remote HANA--> https://github.com/SUSE/ha-sap-terraform-deployments/issues/803
- setting (persistent) DB settings remotely
- query HANA status or information remotely

It needs https://github.com/SUSE/shaptools/pull/69 to be fixed to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Your directory layout should looks like ( all the 3 dirs are in same three dir l
 ```bash
 virtualenv saltvirtenv
 source saltvirtenv/bin/activate
-pip install pyzmq PyYAML pycrypto msgpack-python jinja2 psutil futures tornado pytest-salt mock pytest-cov
+pip install pyzmq PyYAML pycrypto msgpack-python jinja2 psutil futures tornado pytest-salt-factories mock pytest-cov
 pip install -e ../salt
 pip install -e ../shaptools
 rm ../salt/tests/conftest.py # remove this file from the saltstack repo
@@ -115,9 +115,7 @@ salt-call --local state.single saptune.solution_applied "HANA"
 
 ## Dependencies
 
-List of dependencies are specified in the ["Requirements file"](requirements.txt). Items can be installed using pip:
-
-    pip install -r requirements.txt
+A list of dependencies is named above in the `pip install ...` commands.
 
 ## License
 
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -948,21 +948,19 @@ def query(
 
     .. code-block:: bash
 
-        salt '*' hana.query 192.168.10.15 30015 SYSTEM pass
+        salt '*' hana.query 192.168.10.15 30015 SYSTEM pass 'SELECT * FROM SCHEMAS'
     '''
 
     connector = hdb_connector.HdbConnector()
     try:
         connector.connect(host, port, user=user, password=password)
-        rows = connector.query(query)
-        connector.disconnect()
-        # query is returned as a List, so we need to join it into a string
-        result = '\n'.join([str(row) for row in rows.records])
-        return result
+        result = connector.query(query)
 
     except base_connector.QueryError:
         raise exceptions.CommandExecutionError('HANA database query not successfull on {}:{} with query "{}"'.format(
                 host, port, query))
+    finally:
+        connector.disconnect()
 
 
 def reload_hdb_connector():  # pragma: no cover
--- a/salt/modules/hanamod.py
+++ b/salt/modules/hanamod.py
@@ -924,6 +924,47 @@ def wait_for_connection(
             ))
 
 
+def query(
+        host,
+        port,
+        user,
+        password,
+        query):
+    '''
+    Execute a query on a HANA database
+
+    host
+        Host where HANA is running
+    port
+        HANA database port
+    user
+        User to connect to the databse
+    password
+        Password to connect to the database
+    query
+        Query to execute on database
+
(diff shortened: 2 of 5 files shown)


salt/modules/hanamod.py
```diff
@@ -924,6 +924,47 @@ def wait_for_connection(
             ))
 
 
+def query(
+        host,
+        port,
+        user,
```
arbulu89, line 930: One potential improvement would be to give the option to use a stored key instead of the user/password.
But i'm nos sure if this works in pyhdb (in hdbapi certainly works).
Something like this: https://github.com/SUSE/salt-shaptools/blob/79f35d32a9135f956e2c29b886bb57c8ca7470cc/salt/modules/hanamod.py#L678

arbulu89: Indeed, it only works in `hdbapi`. But it would still be insteresting.
The parameter name is `userkey`. Something like `connect(host, post, userkey=userkey, user=user, password=password)`

salt/modules/hanamod.py
```diff
@@ -924,6 +924,47 @@ def wait_for_connection(
...
+    '''
+
+    connector = hdb_connector.HdbConnector()
+    try:
+        connector.connect(host, port, user=user, password=password)
+        rows = connector.query(query)
+        connector.disconnect()
+        # query is returned as a List, so we need to join it into a string
+        result = '\n'.join([str(row) for row in rows.records])
+        return result
```
arbulu89, line 961: I don't know if you will be able to use the return if you run the module directly in salt. I don't think that's possible.
Usually, this return value would be used in a python implemented salt state

salt/modules/hanamod.py
```diff
@@ -924,6 +924,47 @@ def wait_for_connection(
...
+        salt '*' hana.query 192.168.10.15 30015 SYSTEM pass
+    '''
+
+    connector = hdb_connector.HdbConnector()
+    try:
+        connector.connect(host, port, user=user, password=password)
+        rows = connector.query(query)
+        connector.disconnect()
+        # query is returned as a List, so we need to join it into a string
+        result = '\n'.join([str(row) for row in rows.records])
```
arbulu89, line 960: This formatting looks pretty suboptimal. Why not return the result at all and the user can use it openly?

salt/modules/hanamod.py
```diff
@@ -924,6 +924,47 @@ def wait_for_connection(
...
+
+    .. code-block:: bash
+
+        salt '*' hana.query 192.168.10.15 30015 SYSTEM pass
+    '''
+
+    connector = hdb_connector.HdbConnector()
+    try:
+        connector.connect(host, port, user=user, password=password)
+        rows = connector.query(query)
```
arbulu89, line 957: I would call to this `result`

arbulu89 (changes requested):
Hey @yeoldegrove ,
The idea looks good. I have added some comments.
Besides this, we should add a new UT as it looks pretty simple to do.

salt/modules/hanamod.py
yeoldegrove: This would be a completely new feature as the connectors would need to be touched. I opened #85 for this.

salt/modules/hanamod.py
yeoldegrove: You mean getting rid of the `return` completely? Mhm...it seems odd to me to not see the output at all.
Can you elaborate on how this is "better"?

arbulu89: What is this `result` used for? It cannot be used in `salt` directly, as we cannot capture the return, and it is not used anywhere else.
If it's only for logging purpose, just log it.

salt/modules/hanamod.py
arbulu89: You didn't actually do the change here

arbulu89 (changes requested):
@yeoldegrove 
I still see things that should be improved.
- Add some UT
- I don't still like to return a string, only for logging purposes. If we want to log the output, let's do it with good logging code. The code should return the result as it is, so the consumer can use them, and not have a string that is not usable at all

yeoldegrove:
@arbulu89 
- UTs are added
- You are right about the return values, we do not need them. They are already logged inside `INFO` e.g.:
```
[INFO    ] executing sql query: SELECT * FROM SCHEMAS
[INFO    ] query records: [('HANA_XS_BASE', '_SYS_REPO', 'TRUE', datetime.datetime(2022, 1, 19, 7, 52, 40, 623000)), ...
```

README.md
```diff
@@ -115,9 +115,7 @@ salt-call --local state.single saptune.solution_applied "HANA"
 
 ## Dependencies
 
-List of dependencies are specified in the ["Requirements file"](requirements.txt). Items can be installed using pip:
-
-    pip install -r requirements.txt
+A list of dependencies is named avobe in the `pip install ...` commands.
```
arbulu89, line 118: Typo `avobe -> above`

arbulu89 (approved):
Thanks @yeoldegrove ,
Fix the typo and green light

README.md
yeoldegrove: :+1:

yeoldegrove:
@arbulu89 typo is fixed. Please merge.

arbulu89 (approved):
Thanks!

PR #87, opened 2022-03-18
detect cloud aws workaround

yeoldegrove:


salt/states/crmshmod.py
```diff
@@ -596,6 +596,16 @@ def cloud_grains_present(
     __salt__['grains.set']('cloud_provider', cloud_provider)
     changes['cloud_provider'] = cloud_provider
 
+    # workaround for https://github.com/SUSE/ha-sap-terraform-deployments/issues/832
```
arbulu89, line 599: I would say this piece of code should go in the `crm.detect_cloud` module, and not here.
In fact, I don't know why we have the `gcp` part there as well.
This means that the `detect_cloud` module function is kind of broken otherwise

arbulu89 (changes requested):
I would put this abstraction in the module `detect_cloud` function rather than here

salt/states/crmshmod.py
yeoldegrove: The `gcp` part does not change the `cloud_provider` but additional grains `gcp_instance_id/name`.
I guess this is fine...I guess there is just no better place to set this at the moment.

yeoldegrove: @arbulu89 Force pushed an update which moves this to the module.

PR #93, opened 2023-07-11
Salt no longer vendors six (>=salt-3006.0)

yeoldegrove:
Salt no longer vendors six (>=salt-3006.0)
https://github.com/saltstack/salt/issues/63874

- Use shipped library with older salt releases and use packaged version in newer version.

This is related to https://github.com/SUSE/ha-sap-terraform-deployments/issues/908

PR #94, opened 2023-07-14
 ci: fix tests and use SLE containers 

yeoldegrove:
This fixes the currently broken CI pipeline. Github actions/setup-python deprecated python 2.7 https://github.com/actions/setup-python/issues/543 and is not usable anymore.
- The new approach uses real SLE 12 SP5 (python 2.7) and 15 SP5 (python 3.6) container images to run the tests.
- python from official repos is used
- python packages are installed from pip (as most of them are not available in repos)
- codeclimate is removed (as discussed with @arbulu89)

The SLE 15 SP5 tests will fail, as https://github.com/SUSE/salt-shaptools/pull/93 still has to be fixed.


.github/workflows/salt-shaptools-ci.yml
```diff
@@ -11,7 +14,7 @@ env:
 jobs:
   tab:
     name: 'tabspace checking'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
arbulu89, line 17: We should put this in some generic env variable in the top of the file

.github/workflows/salt-shaptools-ci.yml
```diff
@@ -137,53 +144,56 @@ jobs:
...
     strategy:
+      # do not fail if other test fails
+      fail-fast: false
       matrix:
-        os: [ubuntu-20.04]
-        python-version: [2.7, 3.6, 3.7]
+        container:
+          - registry.suse.com/suse/sles12sp5:latest  # python 2.7
```
arbulu89, line 153: We don't we simply use `openSUSE`, we avoid the need of credentials, and in essence the base image would be really similar

.github/workflows/salt-shaptools-ci.yml
```diff
@@ -137,53 +144,56 @@ jobs:
...
-          # on old python version, install and clone ...
-          if test -d /opt/hostedtoolcache/Python/2.7.*; then
-            pip install enum34
-            git clone --branch=openSUSE-3000.3 --depth=50 https://github.com/openSUSE/salt ../salt
+          zypper -n in -y make python
+          if test -f /usr/bin/python3; then
+            # minimal python packages and compilers
+            zypper -n in -y python3-pip python3-devel gcc
+            # use current salt version shipped with SLE 15
+            git clone --branch=openSUSE/release/3006.0 --depth=50 https://github.com/openSUSE/salt ../salt
```
arbulu89, line 174: I didn't remember this hacky thing we had before :sweat_smile: 
If we want to install salt using the same version shipped in SLE, why don't we use zypper?
This way, we would always have the latest rolling version I guess

arbulu89 (approved):
I have to confess that I'm not a big fan of using git cloned versions, as it will become outdated at some point.
But hey, i don't have much suggestion except using zypper/leap, so up to you @yeoldegrove hehe

.github/workflows/salt-shaptools-ci.yml
yeoldegrove: @arbulu89 This is not possible at the moment :(
https://github.com/actions/runner/issues/480

.github/workflows/salt-shaptools-ci.yml
yeoldegrove: @arbulu89 Again, for Leap 42.3 there will be no updated anymore. Also for SLE 12.5 one needs to register an additional Module to install salt.
python-shaptools is also not available for Leap ...

I really would vote to use SLE. One could even think about building images before running the CI steps...

yeoldegrove: What I forgot is that there also is some magic in `tests/run.sh` that uses stuff from the salt repo that is not delivered as rpm.
```
cp salt/modules/*.py ../salt/salt/modules/
cp salt/states/*.py ../salt/salt/states/
cp tests/unit/modules/*.py ../salt/tests/unit/modules/
cp tests/unit/states/*.py ../salt/tests/unit/states/
py.test -vv ../salt/tests/unit/modules/test_hanamod.py ../salt/tests/unit/states/test_hanamod.py  ../salt/tests/unit/modules/test_crmshmod.py  ../salt/tests/unit/modules/test_saptunemod.py ../salt/tests/unit/modules/test_sapcarmod.py ../salt/tests/unit/states/test_crmshmod.py ../salt/tests/unit/modules/test_drbdmod.py ../salt/tests/unit/states/test_drbdmod.py ../salt/tests/unit/states/test_saptunemod.py ../salt/tests/unit/modules/test_netweavermod.py ../salt/tests/unit/states/test_netweavermod.py ../salt/tests/unit/states/test_sapcarmod.py --cov=salt.modules.hanamod --cov=salt.states.hanamod --cov=salt.modules.crmshmod --cov=salt.states.crmshmod --cov=salt.modules.drbdmod --cov=salt.modules.saptunemod --cov=salt.modules.sapcarmod --cov=salt.states.saptunemod --cov=salt.states.drbdmod --cov=salt.modules.netweavermod --cov=salt.states.netweavermod --cov=salt.states.sapcarmod --cov-config .coveragerc --cov-report term --cov-report xml --cov-report html
```

.github/workflows/salt-shaptools-ci.yml
arbulu89: Well, credentials are always a issue hehe
But I understand the reasons as well

arbulu89 (approved):
You have my blessing @yeoldegrove XD
This salt versioning thing was already a pain in the *** hehe

.github/workflows/salt-shaptools-ci.yml
yeoldegrove: I pushed another commit that at least installs the packaged salt version. Had to do a little other change to `tets/run.sh`.

PR #96, opened 2025-09-10
Fix import error salt bundle

arbulu89:
Fix `import crmsh` error when the salt minion is running in a python venv. 
This happens specifially when a salt bundle (venv-salt-minion) is used in the salt client side.

Using the python absolute path guarantees that crmsh will be running the system python installation, which is what we need, as we shouldn't be using the salt bundle python installation in this case.


stefanotorresi (approved):
hackish, but okay :D

